PR #1, opened 2015-05-29
Completed Lunch Attempt for Bank Accounts

MisShellyMac:
One note: letting a user do a check that takes the account to a -10 balance is allowed and then the $2 fee is assessed letting the account total be $-12. This is the only case where it is allowed to be less than the required $10. This is my completed Lunch attempt. Specs did pass as green when I tested. Thanks!


lib/bank_account/account.rb
```diff
@@ -0,0 +1,39 @@
...
+        raise ArgumentError, 'You cannot use a negative number for your initial balance'
+      end
+      @id = id
+      @balance = initial_balance
+    end
+
+
+    def withdraw(amount)
+      if @balance - amount < 0
+        print "This transaction cannot be completed because you don't have enough money."
```
GreyKn, line 18: Instead of `print`, if you use `warn`, it'll let readers know this is a potential problem

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,39 @@
...
+    attr_accessor :id, :balance, :checks
+
+    def initialize (id, initial_balance)
+      @id = id
+      @balance = initial_balance
+      @checks = 3
+    end
+
+    def withdraw(amount)
+      @balance = @balance - amount - 1
```
GreyKn, line 13: I would precheck balance amount so you can't go negative

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,39 @@
+module BankAccount
+  class CheckingAccount < Account
+
+    attr_accessor :id, :balance, :checks
+
+    def initialize (id, initial_balance)
+      @id = id
```
GreyKn, line 7: Because this is mostly the same as the Parent, just with added `@checks`, you can utilize `super` to do less typing.

```
  super
  @checks = 3
end
```

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,39 @@
...
+        return @balance
+      end
+
+      if @checks <= 0
+        @balance = @balance - amount - 2
+        # return @balance
+      else
+        @balance = @balance - amount
+      end
+        @checks = @checks - 1
```
GreyKn, line 29: This line should be shifted left in indent because it's not in a block.

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,41 @@
...
+      @balance = @balance - amount - 2
+      return @balance
+    end
+
+    #This will calculate the interest on the balance and needs to return the amount
+    #of interest and not the new amount with interest. If you call the balance method
+    #it should have the new total with the balance as well. interest needs to be entered
+    #with a 0.XX format.
+    def add_interest(rate)
+      interest = @balance * rate/100
```
GreyKn, line 35: Just as a failsafe, if you do `rate/100.0`, you don't have to worry about rate being an integer.

spec/lib/bank_account/account_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+describe BankAccount::Account do
+  it "responds to the .new class method" do
+    expect(BankAccount::Account).to(respond_to(:new))
+  end
+end
+
+describe "Declining to Open an Account with Less than zero" do
```
GreyKn, line 9: I would shift all the describes to live under your first describe, and change them to `context`.

GreyKn, line 9: You can think about it this way organizationally:

describe : subject
context : action
it : result

spec/lib/bank_account/account_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let (:account) {BankAccount::Account.new(345, -5)}
+  it "Raises an exception" do
+    expect { raise ArgumentError}.to raise_exception (ArgumentError)
```
GreyKn, line 12: What you actually want to do in here is:

```
expect { account }.to raise_error(ArgumentError)
```

spec/lib/bank_account/checking_account_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+describe "Executing a withdraw" do
+  let (:account) {BankAccount::CheckingAccount.new(345, 500,)}
+  it "When withdraw is executed the balance of the bank account will be reduced by withdraw amount and then deduct an additional 1 and lastly will return the new balance" do
+    expect(account.withdraw(50)).to eq(449)
+  end
+end
+
+describe "Executing a withdraw with check" do
+  let (:account) {BankAccount::CheckingAccount.new(345, 500,)}
```
GreyKn, line 17: You left a dangling comma in your argument list. It shouldn't affect your project, but it's messy.

GreyKn:
Hi Michelle,

Good job on the lunch level of this assignment. I liked you tried to limit duplicate code. I left you a comment in one instance where you could get some savings.

The thing with specs is that you can (and often should) have more than 1 `expect`. As you write more specs, think about each element of the test object you want to know about. Like with the `withdraw with check`, you could check both check counts and balance.

-Chef

lib/bank_account/savings_account.rb
davidmfoley, line 35: Minor note: The order of operations here is different from what is implied by the spacing on the line --  it is like:

``` ruby
(@balance * rate) / 100
```

... not like: 

``` ruby
@balance * (rate / 100)
```

PR #2, opened 2015-05-30
Bank Account Breakfast

wangg131:
I finished everything through breakfast. Please let me know if I am repeating myself unnecessarily in my code and rspec. I also accidentally pushed my terminal log so please ignore that. I figure out how to delete it at a later time. Also, more comments and possibly revisions to Bank Account breakfast to come. I just wanted to make the pull request now since I met all the requirements and can refactor later if I have time. Thanks


lib/bank_account/account.rb
```diff
@@ -0,0 +1,37 @@
+module BankAccount
+
+  class Account
+    attr_reader :id, :balance # attributes' default instance method
+
+    #class method
+    def initialize(id, initial_balance)
+      @id = id
+      @balance = initial_balance #local variable
+      # made initial balance equal to instance variable balance, which actually turns out to be the current balance
```
bri-meow: My thoughts are that right now, your code is readable enough that you don't need this comment!  Your variable names are great and make a lot of sense in the context here.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,37 @@
...
+      #this raises an error to prevent the initial balance from being lower than 0
+      unless @balance >= 0
+        raise ArgumentError.new("You can't start out with a negative balance. Add some monies")
+      end
+    end
+
+    def withdraw(amount)
+
+  # If the amount to withdraw is greater than the current balance, it throws and error and returns the original balance
+      unless @balance >= amount
```
bri-meow: I tend to only use `unless` when it makes a ton of sense.  If I was telling someone what this function does (which you did in the comment above), I would use `if` instead.  So something like this instead:

```
if amount >= @balance
  puts "You can't overdraw. Your balance is still $#{@balance}."
```

Also, I'm noticing that this function returns @balance in the first part of the conditional, but the second part doesn't return @balance.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,37 @@
...
+        puts "You can't overdraw. Your balance is still $#{@balance}."
+        return @balance
+
+  # Otherwise, just subtract from the current balance and increment it so balance will always stay current
+      else
+        @balance -= amount
```
brigid2718: Don't need to return @balance here since you are changing it in the function.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+    end
+
+# adds an amount to the current balance and increments it so the balance will always stay current
+    def deposit(amount)
+      @balance += amount
+      return @balance
```
bri-meow, line 34: Don't need to return @balance here since you are changing it in the function.

(Heh, I missed that your spec wants your function to return the current balance when depositing, so nevermind this comment)

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,57 @@
...
+    attr_reader :id, :balance
+
+    def initialize(id, initial_balance)
+      @id = id
+      @balance = initial_balance
+      # again, made initial balance equal to instance variable balance, which actually turns out to be the current balance
+
+      #this raises an ArgumentError to make sure that the initial balance can't be lower than 10
+      unless @balance >= 10
+        raise ArgumentError.new("You can't start out with a balance below $10. Add moar monies.")
```
bri-meow, line 13: moar!!!

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+    end
+
+    def withdraw(amount)
+
+      # I made a new variable to represent the transaction fee added to the amount of withdrawal
```
bri-meow, line 20: Good use of a comment!  Otherwise, you could go without the comment and add a new variable called `fee` and then `amount_with_fee = amount + fee`, but this way is nice and readable too. :)

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,57 @@
...
+      # otherwise for any other withdrawal amount, do a valid withdrawal if the current balance is more than $10
+      # AND if said withdrawal wont make the account go under $10
+      elsif @balance > 10 && ((@balance - amount_with_fee) >= 10)
+        @balance -= amount_with_fee
+
+      # this means that you tried to withdraw an amount that will make savings be less than $10
+      # and so it will just return an error and the unmodified balance
+      else
+        puts "You can't go under $10. Your balance is staying at $#{@balance}."
+        return @balance
```
bri-meow: The directions for the assignment aren't very clear here, but I would want the @balance to be returned no matter what (not just in the case that your account goes under $10), so I would move the return statement outside of the conditional.

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    end
+
+    # inherited the deposit behavior from the parent class, BankAccount::Account
+    def deposit(amount)
+      super
```
bri-meow, line 43: sweet usage of super here!  I like that you left the comment explaining it, but that's because I never remember what super does.

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,49 @@
+require './lib/bank_account/account'
+require 'spec_helper.rb'
+# root folder do rspec..
+describe BankAccount::Account do
+  let (:new_account200) {BankAccount::Account.new(100, 200)}
```
bri-meow, line 5: I like the use of the let statement, keeps your tests nice and DRY!

You could maybe afford to give it an even more descriptive name, even if it's a little bit long.

bri-meow:
I think your code for Breakfast looks pretty good!  Your code is pretty DRY and readable.  I like that you used the let bit of rspec to create a bank account object to use.

I think the only thing to watch for is that within your functions you make sure they have consistency about what they return.  I noted some of this inconsistency in your functions that return @balance in some parts of the conditional, but not in others.

Otherwise, good job!  Good use of comments and variable naming for readability!

lib/bank_account/account.rb
wangg131: Thank you! Looking back at it now, I definitely wasn't consistent enough.

lib/bank_account/savings_account.rb
wangg131, line 13: hahah yes, u can hazz moar

lib/bank_account/account.rb
wangg131: Thank you! Looking back at it now, I definitely wasn't consistent enough.

wangg131:
Thank you for grading this and letting me know about the inconsistencies of returning or not returning @balance!

PR #3, opened 2015-05-30
Completed bank accounts project

sallyamoore:
I completed the Lunch tier on this project. I learned a lot about inheritance in the process! It would be useful to have a little input on my specs... still getting used to writing them, and I have a feeling I could be more concise. Also, do I have too many or too few specs?


lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
+module BankAccounts
+
+  class Account
+    attr_accessor :id, :min_balance, :initial_balance, :balance, :fee
```
GreyKn: Great use of inheritance to define accessors and member variables. :100: 

I do have a question though: Why have accessors defined for `min_balance`, `initial_balance`, and `fee`? Will the users of the classes need access to those attributes?

GreyKn: Oh, and given the context, `attr_accessor` gives both read and write access to the attribute. You may want to use `attr_reader` to users can't arbitrarily change the attributes from outside

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    def initialize(id, initial_balance = 0)
```
GreyKn, line 6: Cool, default arguments! :clap:

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+      @min_balance = 0
+      @initial_balance = initial_balance
+      # Raises ArgumentError if initial_balance < 0
+      check_initial_balance(initial_balance)
+      @balance = initial_balance
+      @fee = 0
+    end
+
+    def check_initial_balance(initial_balance)
+      if initial_balance == nil || initial_balance < @min_balance
```
GreyKn: Good, you checked for nil values :+1: 

Let me show you another way to check for nils in ruby:

```
if initial_balance.nil? || initial_balance < @min_balance
```

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+      # warn if there's a fee
+      warn("Warning: This transaction will incur a $#{fee} fee.") if @fee > 0
+
+      # does not allow the account to go negative.
+      # outputs a warning message & returns the original un-modified balance
+      if new_balance < @min_balance
+        warn("Warning: Insufficient funds! You cannot withdraw #{amount}. Your account only has #{@balance}.")
+      else
+        # updates account balance
+        @balance = new_balance.round(2)
```
GreyKn: Good, you're accounting for fractional cents. But normally, you only want to do rounding when you display the number, internally you want to keep those fractions. 

Unless you're doing this: https://youtu.be/GlRG9x0JRMc?t=2m4s... then we've never had this conversation...

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+    def check_initial_balance(initial_balance)
+      if initial_balance == nil || initial_balance < @min_balance
+        raise ArgumentError.new("Initial balance must be #{min_balance} or greater.")
+      end
+    end
+
+    # Subtracts amount from balance and returns balance
+    def withdraw(amount)
+      # amount and fee are subtracted from balance.
+      new_balance = @balance - amount - @fee
```
GreyKn, line 25: Great, you're prechecking the value before writing it to `@balance` :+1:

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      # warn if there's a fee
+      warn("Warning: This transaction will incur a $#{fee} fee.") if @fee > 0
```
GreyKn, line 28: I like that you're using `warn` for these messages. It means we can redirect those messages to a separate stream for processing.

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      # does not allow the account to go negative.
+      # outputs a warning message & returns the original un-modified balance
+      if new_balance < @min_balance
+        warn("Warning: Insufficient funds! You cannot withdraw #{amount}. Your account only has #{@balance}.")
```
GreyKn, line 33: You're providing very specific messaging in your warning, telling people why the error is happening :100:

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    # Adds amount to balance and returns balance
+    def deposit(amount)
+      @balance += amount
+      return @balance.round(2)
+    end
+
+    # returns balance
+    def balance
+      @balance.round(2)
```
GreyKn, line 49: Considering you're rounding in both your `withdraw` and `deposit` methods, do you need to use `round` here?

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    # Adds amount to balance and returns balance
+    def deposit(amount)
+      @balance += amount
+      return @balance.round(2)
+    end
+
+    # returns balance
+    def balance
```
GreyKn, line 48: Interesting. You're defining both an attribute accessor and a method called `balance`. When a user calls that method, which one are they triggering?

lib/bankaccounts/checking-account.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    def initialize(id, initial_balance)
+      super
+      @min_balance = 0
+      check_initial_balance(initial_balance)
+      @fee = 1
+      @checks_used = 0
+    end
+
+    # withdraw is inherited from Account; therefore, it does not appear here.
```
GreyKn, line 14: Good comment on why `#withdraw` isn't explicitly defined

lib/bankaccounts/checking-account.rb
```diff
@@ -0,0 +1,48 @@
...
+    # withdraw(amount) subtracts amount from balance, incurs $1 fee.
+    # if account would fall below @min_balance, outputs warning & returns unmodified balance.
+
+    def withdraw_using_check(amount)
+      # re-assigns @min_balance to allow overdraft up to -$10
+      @min_balance = -10
```
GreyKn: Ah, be careful. You're reassigning both `@min_balance` and `@fee`, so now if someone uses `#withdraw_using_check` first, then `#withdraw`, you'd end up with incorrect conditionals happening.

What you probably want to do is use local variables that exist only in the context of `#withdraw_using_check`, so it doesn't affect `#withdraw`

lib/bankaccounts/checking-account.rb
```diff
@@ -0,0 +1,48 @@
...
+      # increments @checks_used
+      @checks_used += 1
+
+      # User can have three free checks in 1 month, any subsequent checks incur $2 fee
+      @checks_used <= 3 ? @fee = 0 : @fee = 2
+
+      # Subtracts amount
+      new_balance = @balance - amount - @fee
```
GreyKn: You're duplicating some code here from `#withdraw`, can you think of a way to use what you're already written inside this method?

lib/bankaccounts/checking-account.rb
```diff
@@ -0,0 +1,48 @@
...
+    # if account would fall below @min_balance, outputs warning & returns unmodified balance.
+
+    def withdraw_using_check(amount)
+      # re-assigns @min_balance to allow overdraft up to -$10
+      @min_balance = -10
+      # increments @checks_used
+      @checks_used += 1
+
+      # User can have three free checks in 1 month, any subsequent checks incur $2 fee
+      @checks_used <= 3 ? @fee = 0 : @fee = 2
```
GreyKn: Good use of the ternary operator

lib/bankaccounts/savings-account.rb
```diff
@@ -0,0 +1,27 @@
...
+    # withdraw is inherited from Account; therefore, it does not appear here.
+    # withdraw(amount) subtracts amount from balance, incurs $2 fee.
+    # if account would fall below @min_balance, outputs warning & returns unmodified balance.
+
+    # Calculates interest and adds it to the balance.
+    def add_interest(rate)
+      # Calculates interest & updates balance
+      interest = @balance * (rate/100)
+      @balance += interest
+      # Returns calculated interest _not_ the updated balance
```
GreyKn, line 21: Great comment. This is inconsistent with the other methods (everything else returns the modified balance), so having a comment here clarifies things a lot.

lib/bankaccounts/checking-account.rb
```diff
@@ -0,0 +1,48 @@
+module BankAccounts
+
+  # CheckingAccount inherits from Account class
+  class CheckingAccount < Account
+    attr_accessor :checks_used
+
+    def initialize(id, initial_balance)
+      super
+      @min_balance = 0
+      check_initial_balance(initial_balance)
```
GreyKn: `#check_initial_balance` is called on `#super` and again here. Can you think of a way of changing the order in which `#super` is called so it happens only once?

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'spec_helper'
+
+describe BankAccounts::Account do
+
+  describe "class methods" do
+    it "responds to 'new'" do
```
GreyKn: Technically, you shouldn't need this test as all classes inherit from Object, which has the new method. What you could check is whether or not calling `BankAccounts::Account.new(x,y)` creates an Account object will the right attributes.

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      expect(BankAccounts::Account).to respond_to :new
+    end
+
+    let(:account) do
+      BankAccounts::Account.new("Voltron", 1000)
+    end
+
+      it "subtracts withdraw(amount) and returns updated balance" do
```
GreyKn: This `it` needs to be shifted left one indent

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        expect(account.withdraw(50)).to eq(950.00)
+      end
+
+    context "when withdraw(amount) would result in a negative value of balance" do
+      let(:account) do
```
GreyKn, line 15: You shouldn't need to redefine `account` here. It will inherit it from the parent context

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+          expect(account.withdraw(5032)).to eq(1000)
+        end
+
+        it "raises a warning" do
+          expect { account.withdraw(2000) }.to output("Warning: Insufficient funds! You cannot withdraw 2000. Your account only has 1000.\n").to_stderr
+        end
+    end
+
+    describe "#deposit(amount)" do
+    let(:account) do
```
GreyKn: This `let` needs to be shifted to the right

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,53 @@
+module BankAccounts
+
+  class Account
+    attr_accessor :id, :min_balance, :initial_balance, :balance, :fee
+
+    def initialize(id, initial_balance = 0)
+      @id = id
+      @min_balance = 0
+      @initial_balance = initial_balance
```
GreyKn, line 9: Does this class need `@initial_balance`? It doesn't seem to be used anywhere.

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'spec_helper'
+
+describe BankAccounts::Account do
+
+  describe "class methods" do
+    it "responds to 'new'" do
+      expect(BankAccounts::Account).to respond_to :new
+    end
+
+    let(:account) do
```
GreyKn, line 6: For shorter statements, you could write this as:

```
let(:account) { BankAccounts::Account.new("Voltron", 1000) }
```

spec/lib/checking-account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      let(:checking_account) do
+        BankAccounts::CheckingAccount.new("Voltron2", 1002)
+      end
+
+        it "returns original balance" do
+          expect(checking_account.withdraw(5032)).to eq(1002)
+        end
+
+        it "allows withdrawals betwee 0 and 10" do
+          expect(checking_account.withdraw(1000)).to eq(1)
```
GreyKn, line 35: What is this spec supposed to test? The description is not clear. It says something about withdrawing 0 and 10, but the test is withdrawing 1000?

spec/lib/checking-account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+        expect(checking_account.withdraw_using_check(100)).to eq(87)
+      end
+
+      it "responds to 'reset_checks'" do
+        expect(checking_account).to respond_to :reset_checks
+      end
+
+      context "after the third withdraw_using_check call" do
+        it "charges a $2 fee" do
+          expect(checking_account.withdraw_using_check(7)).to eq(80)
```
GreyKn: Good test. You could use some arrays and a loop to reduce the amount of code repetition

```
amounts = [7, 2, 7, 10, 7]
expected_balance = [80, 78, 71, 59, 50]
amounts.each_index do |index|
   expect(checking_account.withdraw_using_check(amounts[index])).to eq(expected_balance[index])
end

```

spec/lib/savings-account_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        it "raises an ArgumentError" do
+          expect { BankAccounts::SavingsAccount.new(898776, nil) }.to raise_exception(ArgumentError)
+        end
+      end
+    end
+
+    describe "unique responses to inherited methods" do
+      describe "raise_exception when initial_balance is < 10 and > 0" do
+        it "raises an ArgumentError" do
+          expect { BankAccounts::SavingsAccount.new(898776, 5) }.to raise_exception(ArgumentError)
```
GreyKn, line 30: One thing you can do is test every condition using a loop, instead of relying on one number inside the range, like this:

```
(1...10).each do |initial_deposit|
  expect { BankAccounts::SavingsAccount.new(898776, initial_deposit) }.to raise_exception(ArgumentError)
end
```

GreyKn:
Hi Sally,

Great work on the lunch level of this assignment. You had several creative solutions in handling class inheritance which I enjoyed looking over. I don't necessarily think you have too many specs, but I think a couple could be written in a way to cover more scenarios with less typing.

It's a little nitpicky, but please try to remain consistent with your indenting. I got derailed on occasion when I came across unexpected indenting.

Let me know if you have any questions.
-Chef

sallyamoore:
This is all SO HELPFUL!!! Thank you for providing a review that was both thorough and reinforcing. :) I hope to incorporate your suggestions and push a new commit today or tomorrow.

lib/bankaccounts/account.rb
sallyamoore, line 49: since `balance` is a separate method that could be called before someone withdraws or deposits, I thought it should be here too...

lib/bankaccounts/account.rb
sallyamoore, line 48: They should return the same thing given what I've written, right? How might this mess things up? Part of the assignment was to create a `balance` method; otherwise, I could've just left it as a readable attribute, I think...

lib/bankaccounts/checking-account.rb
sallyamoore: Ahhh! That makes so much sense! :)

lib/bankaccounts/account.rb
GreyKn, line 48: I think the last thing defined will take precedence. So in this case, you defined the balance method last, so it's what's triggered when you call `#balance`. Considering this, you shouldn't need to define the attr_accessor version.

lib/bankaccounts/checking-account.rb
sallyamoore: It looks like I don't need to call it outside of the call within `#super`. I removed it, and all of my specs still pass. Will this break something I'm not seeing?

spec/lib/account_spec.rb
sallyamoore, line 6: Heh, I had it that way and another TA suggested that I use do/end because the lines were so long. :) I gather that either way is acceptable...

sallyamoore:
I just pushed a new commit. Take a look if you have time. Thanks again for your help!

spec/lib/checking-account_spec.rb
sallyamoore, line 35: Fixed!

spec/lib/account_spec.rb
sallyamoore: fixed

spec/lib/account_spec.rb
sallyamoore, line 15: Ah yes...

lib/bankaccounts/account.rb
GreyKn, line 49: Good reasoning

lib/bankaccounts/checking-account.rb
GreyKn: In this case, no, because min_balance is 0 for both CheckingAccount and Account. It would cause it to break if you removed it from SavingsAccount though.

spec/lib/account_spec.rb
GreyKn, line 6: Yep. 'Too long' is subjective, so either way is fine.

PR #4, opened 2015-05-30
Completed BankAccounts -- Through Dinner Level

lo-k:
Here is my completed BankAccounts project through the Dinner level requirements.

Thanks for your feedback!


roseaboveit:
I'm starting this review. Going to work on it today and tomorrow

spec/account_spec.rb
```diff
@@ -0,0 +1,98 @@
+require "spec_helper"
+require "bankaccount"
+
+describe BankAccount::Account do
+	
+	it "responds to class method called new" do
+		expect(BankAccount::Account).to respond_to :new
+	end
+
+	describe "it creates an instance of a new Account with an id and a positive initial_balance" do
```
roseaboveit, line 10: When you run your specs, errors will be displayed as the `describe` strings followed by the `it` string.
As such the `describe` strings should be short and descriptive of all the specs nested within it.
One example that might work here is "Initialized account"
Then everything in that section should test that the initialized account and the initialization process is as you think it should be.
`it` strings should let you know what the problem is before you go and look at the code.
What we are testing here is that it sets things if it is valid and raising errors if invalid.
As such your 3rd test in this describe block has a very good description.
The other two are testing the general concept of setting the id and the initial balance. The exact value is part of the test itself but not really the display string.

While looking for helpful resources that don't eat up a ton of time I found this: http://jakegoulding.com/presentations/rspec-structure/#slide-0 Feel free to tag me here or on twitter if you want to talk more about this concept. It might not be something that was covered in much depth yet.

spec/account_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+			it "responds to the method 'withdraw'" do
+				expect(account).to respond_to :withdraw
+			end
+
+			context "reduces the account balance when the withdraw method is called" do
+				it "returns 4000 when you withdraw 1000" do
+					expect(account.withdraw(1000)).to eq 4000
+				end
+
+				it "returns 3500 when you withdraw 1000 and then withdraw 500" do
```
roseaboveit, line 39: It is not immediately obvious why you would want this test as the concept is covered by the previous test. Is it testing that you can chain the method? Or is it testing that the return state of withdraw changes the balance in a way that persists? This would be good to clarify in the it string. If it is the chaining, that falls into the category of things we don't need to test because we can trust the ruby core team to have tested core ruby functionality. The persistence bit might be worth testing if the implementation changed and no longer persisted you'd want your tests to alert you about it.

spec/account_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+				it "returns the initial balance (5000) if you have not made any transactions" do
+					expect(account.balance).to eq 5000
+				end
+
+				it "returns 4000 as current balance after you've withdrawn 1000" do
+					account.withdraw(1000)
+					expect(account.balance).to eq 4000
+				end
+
+				it "returns 2525 as current balance after you've -3000 +1000 -500 +25" do
```
roseaboveit, line 85: Regarding my last comment. I thought of another possibility. And that is an integration test. The other tests are very much unit tests which is most of what we use RSpec for. Integration tests involving using several methods can be quite valuable as well. I feel like you only need one for this class and this one is good because it does several transactions that could simulate actual user data.

lib/bankaccount/checking_account.rb
```diff
@@ -0,0 +1,59 @@
...
+
+
+		def deposit(amount)
+			super
+		end
+
+		def balance
+			super
+		end
+
```
roseaboveit, line 56: The way that inheritance works, if there are no changes you want to make to the parent method, you don't even need to state it. This class already knows about those methods.

lib/bankaccount/account.rb
```diff
@@ -0,0 +1,37 @@
+module BankAccount
+
+	class Account
+
+		attr_accessor :id, :initial_balance, :account_balance
```
roseaboveit, line 5: Did you have a reason behind making these variables accessible?

lib/bankaccount/account.rb
```diff
@@ -0,0 +1,37 @@
...
+
+		def initialize(id, initial_balance)
+			raise ArgumentError.new "Please enter a positive initial balance." unless initial_balance >= 0
+				@id = id
+				@initial_balance = initial_balance
+				@account_balance = initial_balance
+		end
+
+		def withdraw(amount)
```
roseaboveit, line 14: I like this method. One possibility for improvement might be to start by checking for the error condition first and only setting the variable to a new value if it passes the condition. It would save you a line or two.

lib/bankaccount/checking_account.rb
```diff
@@ -0,0 +1,59 @@
...
+				@account_balance = @account_balance + amount + fee
+				raise ArgumentError.new "Insufficient funds to make this withdrawl including a $1 transaction fee. Your current balance is $#{@account_balance}."
+				return @account_balance
+			else 
+				return @account_balance
+			end
+		end
+
+		def withdraw_using_check(amount)
+			fee = (@free_checks > 0) ? 0 : 2
```
roseaboveit, line 29: ooOOOooo nice ternery operator!!!

lib/bankaccount/checking_account.rb
```diff
@@ -0,0 +1,59 @@
+module BankAccount
+
+	class CheckingAccount < Account
+
+		attr_reader :id, :initial_balance, :account_balance, :free_checks
+
+		def initialize(id, initial_balance)
+			raise ArgumentError.new "Please enter a positive initial balance." unless initial_balance > 0
```
roseaboveit, line 8: I like this line. I know you've been taught to do ArgumentErrors but I hadn't at your point in the program so they all seem super shiny and advanced. Also using the `unless` afterward is super slick!

spec/checking_account_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+			context "you can withdraw money until the account balance is < -10" do
+				it "returns -5 when you withdraw 5005" do
+					expect(checking_account.withdraw_using_check(5005)).to eq -5
+				end
+				it "returns an ArgumentError if you withdraw 5011" do
+					expect{ checking_account.withdraw_using_check(5011) }.to raise_exception(ArgumentError)
+				end
+			end
+
+			context "you get 3 free checks to withdraw from your account for free. After that, there is a 2 fee per withdrawal." do
```
roseaboveit, line 66: I love this set of tests. It covers the whole "free check" feature and "checking withdrawl" feature!

lib/bankaccount/money_market_account.rb
```diff
@@ -0,0 +1,91 @@
+module BankAccount
+
+	class MoneyMarketAccount < Account
+
+		attr_reader :id, :initial_balance, :account_balance, :frozen_account, :allowed_transactions
+
+		def initialize(id, initial_balance)
+			raise ArgumentError.new "Please enter an initial balance of at least $10,000." unless initial_balance >= 10_000
```
roseaboveit, line 8: You taught me something new! I didn't know you could put an underscore for helping to dilineate place value in integers. That makes it really readable!

lib/bankaccount/money_market_account.rb
```diff
@@ -0,0 +1,91 @@
...
+			@account_balance += interest.round(2)
+			return interest.round(2)
+		end
+
+		# reset_transactions -- I did this method the opposite way 
+		# as the instructions said since my logic is counting down 
+		# transactions from 6 rather than adding them up to 6
+		def reset_transactions 
+			@allowed_transactions = 6
+			return @allowed_transactions
```
roseaboveit, line 87: Ruby has implicit return. What this means is that in most cases the last object used in the method is the return value of the method. Try removing line 87 and trying it out for yourself. This can get weird with loops and if statements so if it is complicated definitely use an explicit return.

lib/bankaccount/money_market_account.rb
```diff
@@ -0,0 +1,91 @@
...
+				else
+					return @account_balance
+				end
+			end
+
+		end
+
+
+		def deposit(amount)
+			deposit_needed = 10_000 - @account_balance
```
roseaboveit, line 46: This method was a great choice to use your commenting skills!

spec/checking_account_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+				expect(checking_account.balance).to eq 3508
+			end
+		end
+
+		describe "#withdraw method" do
+			context "you can withdraw money from your CheckingAccount for a 1 transaction fee, unless it would make your account balance go below 0" do
+				it "returns 4499 if you withdraw 500 and it has a 1 transaction fee" do
+					expect(checking_account.withdraw(500)).to eq 4499
+				end
+				it "raises an ArgumentError if you try to withdraw 6000" do
```
roseaboveit, line 50: Nice job remembering to test the sad path!

spec/money_market_account_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+		end
+
+		it "will not deposit additional money if you have no more allowed_transactions" do
+			6.times do
+				money_market_account.deposit(100)
+			end
+			expect(money_market_account.deposit(400)).to eq 15_600
+		end
+
+		# context "MoneyMarketAccounts get 'frozen' if their balance is < 10,000" do
```
roseaboveit, line 39: Nicely done attempting to test the frozen feature. It is definitely tricky. There are a couple things to consider that might help you on your quest:
Read through how you implemented the feature and remember that each line gets executed in order. Where does an account get frozen? Where does the subtraction happen? Make sure your test reflects your implementation. 
Well done with the attempt though. Recognizing that something should be tested is the first step and one that is hard for many.

roseaboveit:
You did a really great job on this project.
There were so many new things and so many recently learned things that you had to bring together and you did so with style.
Things I particularly loved:
- All the tests. You wrote so many tests. This is AWESOME!
- All the commits. You committed early and often. This is a great habit to get into.
- You did a great job weaving together all the pieces of knowledge you have to bring your best work to this assignment.

Things I'd love to see next time:
- So, I think your environment is set up to use tabs instead of spaces and/or using 4 instead of 2. It actually makes your code harder to read. In ruby we use 2 spaces as the convention. Different languages have different conventions so depending on what languages you use in your career you may have different settings in the future.
- The other comments that I made I'm unsure whether they are things you have heard before. If they are, look into them and try to keep them in mind going forward. If not, don't even worry about it, it will come up eventually and so I guess I just gave you a taste of the future.

Like I said, I liked this project a lot. If you would like to talk to me about any of this feel free to tag me here or poke me on twitter. You can also poke me in class I'm usually in on Wednesday afternoons.

lo-k:
Thank you so much for your feedback, @roseaboveit ! Your very thorough notes are extremely helpful. Your points about implicit returns, inheritance, and more mindful attr accessors will save me several lines of code and make it more efficient. I also do acknowledge that I need practice writing more succinct rspec tests. I will definitely keep your notes in mind as I move forward with other projects!

PR #5, opened 2015-05-31
BankAccounts - Lunch level 

EliaMG:
Completed lunch. Feel pretty good about my specs, feel like there may be some opportunities for code refactor as I'm not sure it is as DRY as it could be.


lib/classes/account_class.rb
```diff
@@ -0,0 +1,36 @@
+module BankAccount
+
+  class Account
+    # changed to a reader since it doesn't seem like these attributes should be able to be overwritten
```
dougcole, line 4: :+1: removing the ability of external code modifying instance variables generally leads to better encapsulation and generally better object oriented design.

lib/classes/account_class.rb
```diff
@@ -0,0 +1,36 @@
...
+    attr_reader :id, :initial_balance, :current_balance
+
+    def initialize(id, initial_balance)
+      raise ArgumentError.new ("You cannot create an account with a negative balance") \
+        unless initial_balance > 0
```
dougcole, line 9: If you need to wrap onto a second line I'd just go ahead and use the full unless/end structure rather than the postfix unless, as I think it's clearer.

lib/classes/account_class.rb
```diff
@@ -0,0 +1,36 @@
...
+      else
+        @current_balance = @current_balance - amount
+      end
+    end
+
+    def deposit(amount)
+      @current_balance = @current_balance + amount
+    end
+
+    # seems a little redundant with an attr_reader that can call current_balance
```
dougcole, line 31: agreed!

lib/classes/checkingaccount_class.rb
```diff
@@ -0,0 +1,47 @@
+module BankAccount
+
+  class CheckingAccount < Account
+    attr_reader :checks
+
+    #inherits passed in parameters from Account class
+    def initialize(id, initial_balance)
+    super(id, initial_balance)
```
dougcole, line 8: You don't actually need to pass in arguments to `super` if they are the same as the current method

lib/classes/checkingaccount_class.rb
```diff
@@ -0,0 +1,47 @@
...
+    @current_balance = initial_balance
+    @checks = 0
+    end
+
+    # used super to inherit functionality for overdraft
+    # may not be the best use of super since still needed the if/else block for the transaction fee
+    # could potentially refactor Account #withdraw to use a variable that accounts for fees/minimum balances
+    # but seems like it may be a little more complicated than needed
```
dougcole, line 16: I like the idea of sharing code via super, but in this case, the overdraft functionality is actually a little different between the two classes as in the checking case you need to protect against `@current_balance > amount + 1`.

lib/classes/checkingaccount_class.rb
```diff
@@ -0,0 +1,47 @@
...
+      # controls for overdraft of up to $10
+      if amount > @current_balance + 10
+        puts "$#{amount} is too large of of a withdrawal. Your current balance is $#{@current_balance}."
+        return @current_balance
+      # 3 free check withdrawals
+      elsif @checks < 3
+      @current_balance = @current_balance - amount
+      # once @checks > 3, charges $2 transaction fee
+      else
+      @current_balance = @current_balance - amount - 2
```
dougcole, line 36: With the extra $2 fee I think you need to recheck for overdrafts here, something more like:
 `amount > @current_balance + 10 - 2`

lib/classes/checkingaccount_class.rb
dougcole, line 16: There are a few other approaches you could take to sharing code. For example if you make a method

``` ruby
def fee
 0
end
```

On the base `Account` class and the rewrite the `withdraw` conditional like: `if amount > @current_balance - fee` you could override the `fee` method here and completely reuse the withdraw method. That probably isn't worth it in this context, but hopefully it helps show other ways to share code.

spec/account_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it "raises an ArgumentError if a negative account balance is initialized" do
+      expect{BankAccount::Account.new(3, -5000)}.to raise_error(ArgumentError)
+    end
+  end
+
+  describe "It has instance methods with functionality" do
+    let (:bankaccount) {BankAccount::Account.new(2, 10000)}
+
+    it "Returns a balance of 5000 when 5000 is withdrawn" do
+      expect(bankaccount.withdraw(5000)).to eq(5000)
```
dougcole, line 27: This is very much a personal preference, but I think it's worth duplicating the: 
`bankaccount = BankAccount::Account.new(2, 10000)`
line here, rather than in the `let` as it makes it much easier to see the math that is being done. Otherwise figuring out that the math being done is 10000 - 5000 = 5000 is split by several lines of code.

dougcole:
Awesome work! Let me know if any of my feedback didn't make sense or if you have any questions for me.

lib/classes/checkingaccount_class.rb
EliaMG, line 16: Thanks for this thought process write out! I wasn't sure if I was being completely redundant so I'm glad I wasn't way too off base. I appreciate your review!

PR #6, opened 2015-05-31
dinner version

sojeri:
Here is my dinner version. I had some fun with the flavor text at the end. >_>


lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    attr_reader :balance, :id
+
+
+    def initialize(id, initial_balance)
+      if initial_balance >= MINIMUM_BALANCE
+        @balance = 0
+        update_balance(initial_balance.round(2))
+      else
+        raise ArgumentError.new("An account cannot be created with a negative balance.")
```
ktregina: You could also do this like so: 

```
  raise ArgumentError.new("An account cannot ... ") if initial_balance < MINIMUM_BALANCE

  @balance = 0
  update_balance(initial_balance.round(2))
```

Since `raise` causes the method to exit, the rest of the method will not be executed if the error is raised, so you can get rid of the whole if/else block. This pattern is called a guard clause. The way you did it is totally fine (and does exactly the same thing), this structure is just a little bit more readable.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      if (@balance >= RECOMMEND_MONEY_MARKET) && (@balance < RECOMMEND_SISTER_BANK) && (self.class != MoneyMarketAccount)
+        puts MONEY_MARKET_RECOMMENDATION
+      end
+
+      if (@balance >= RECOMMEND_SISTER_BANK)
+        raise ArgumentError.new(SISTER_BANK_RECOMMENDATION)
+      end
+
```
ktregina: Similar comment here: you could add the line `raise ArgumentError.new(SISTER_BANK_RECOMMENDATION) if initial_balance >= RECOMMEND_SISTER_BANK` to the top of the `initialize` method, and then you could get rid of the `if` block that raises it, and also get rid of the middle conditional (`@balance < RECOMMEND_SISTER_BANK`) in the previous `if` block, since the method would already have raised an error if that condition wasn't met. It can get hard to read a series of a lot of conditionals, so it's good to simplify them when you can.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,125 @@
...
+      return true
+    end
+
+
+    def validate_number(input)
+      unless (input.class == Fixnum) || (input.class == Float)
+        raise ArgumentError.new("That is not a number!")
+      end
+
+      return true
```
ktregina: Since this method's purpose is essentially "throw an error or do nothing", it doesn't really need to be used in conditionals, so it doesn't really need to return `true` (or anything). You can just call it on its own when you need it, rather than put it in a conditional, because you know that if it doesn't throw an error, the validation has passed and the program can continue to run.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+
+    def validate_withdrawal(amount)
```
katiespiders: E.g. here. You could just get rid of the `unless` block and call `validate_number(amount)`. Ruby will actually never get into this `unless` block at all to return `false`, because `validate_number` never returns `false`, only throws an error.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,125 @@
...
+      unless validate_number(amount)
+        return false
+      end
+
```
katiespiders: E.g. here. You could just get rid of the `unless` block and call `validate_number(amount)`. Ruby will actually never get into this `unless` block at all to return `false`, because `validate_number` never returns `false`, only throws an error.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,125 @@
...
+      return true
+    end
+
+
+    def validate_withdrawal(amount)
+      unless validate_number(amount)
+        return false
+      end
```
ktregina: E.g. here. You can just get rid of the `unless` block and call `validate_number(amount)` on its own. The program will actually never get into this `unless` block at all to return `false`, because `validate_number` never returns `false`, only throws an error.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      return @balance
+    end
+
+
+    def deposit(amount)
+      # deposits can only be numeric values.
+      unless validate_number(amount)
+        return @balance
+      end
```
ktregina: (see comments on [line 108](https://github.com/Ada-Developers-Academy/C3Projects--BankAccounts/pull/6/files#diff-0a908e2d048cf4f37a010dcdb481267bR108), [line 115](https://github.com/Ada-Developers-Academy/C3Projects--BankAccounts/pull/6/files#diff-0a908e2d048cf4f37a010dcdb481267bR115))

lib/bank_accounts/savings_account.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    # create a new savings account.
+    def initialize(id, initial_balance)
+      if initial_balance < MINIMUM_BALANCE
+        raise ArgumentError.new("You cannot create a savings account with less than a $#{ MINIMUM_BALANCE } initial deposit.")
+      end
+
+      super(id, initial_balance)
+
+      return true
```
ktregina: `initialize` doesn't need to return anything. It's never (that I've seen, anyway, although I'm still kind of a noob (hi from cohort[1]!)) called explicitly, it's only called by the `new` method on classes, which ignores anything returned by `initialize`.

lib/bank_accounts/savings_account.rb
```diff
@@ -0,0 +1,81 @@
...
+        # add interest to balance.
+        update_balance(interest)
+
+        # return interest.
+        return interest
+      else
+        return false
+      end
+    end
+
```
ktregina, line 42: Here too you could get rid of the whole if/else by calling `validate_number` at the top as a guard clause. There's no way for `validate_number` to return false, so there's no way for this method to return `false` either.

lib/bank_accounts/savings_account.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    # withdraw money from account.
+    def withdraw(amount)
+      if super(amount + WITHDRAWAL_FEE)
+        return @balance
+      else
+        return @balance
+      end
+    end
```
ktregina, line 49: I think that you should be able to write this method as a one-liner: `super(amount + WITHDRAWAL_FEE)`. You shouldn't need either the conditional or the explicit return (the parent `withdraw` method always returns `@balance`, so this will also always return `@balance`.) I'm not 100% sure about this though cause I am still kind of a noob (hi from cohort[1]!)

lib/bank_accounts/savings_account.rb
```diff
@@ -0,0 +1,81 @@
...
+    private
+    ##--------------------------------------------------------------------------
+    # PRIVATE METHODS
+
+
+    # check withdrawal amount is valid (not below minimum balance).
+    def validate_withdrawal(amount)
+      unless validate_number(amount)
+        return false
+      end
```
ktregina: Same deal w/ guard clause here

lib/bank_accounts/checking_account.rb
```diff
@@ -0,0 +1,100 @@
...
+    attr_reader :checks_used
+
+
+    def initialize(id, initial_balance)
+      super(id, initial_balance)
+
+      @checks_used = 0
+
+      return true
+    end
```
ktregina, line 25: `initialize` doesn't need to return a value (explained [here](https://github.com/Ada-Developers-Academy/C3Projects--BankAccounts/pull/6/files#diff-5671ee19f228f2b2c21bd3392819282bR25) -- sorry I'm jumping around)

lib/bank_accounts/checking_account.rb
```diff
@@ -0,0 +1,100 @@
...
+
+
+    # withdraw money from account.
+    def withdraw(amount)
+      if super(amount + WITHDRAWAL_FEE)
+        return @balance
+      else
+        return @balance
+      end
+    end
```
ktregina, line 40: See [this comment](https://github.com/Ada-Developers-Academy/C3Projects--BankAccounts/pull/6/files#diff-5671ee19f228f2b2c21bd3392819282bR56) (sorry again for the jumping!)

lib/bank_accounts/checking_account.rb
```diff
@@ -0,0 +1,100 @@
...
+      return true
+    end
+
+
+    def validate_check_withdrawal(amount)
+      if (@checks_used + 1) > 3
+        amount += EXCESS_CHECKS_FEE
+      end
+
+      if validate_withdrawal(amount)
```
katiespiders: This method can just end with `validate_withdrawal(amount)`.

lib/bank_accounts/checking_account.rb
```diff
@@ -0,0 +1,100 @@
...
+        return true
+      end
+
+      return false
```
ktregina, line 84: This method can just end with `validate_withdrawal(amount)`, no need for the conditional (or explicit return).

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+        @transactions = 0
+
+        super(id, initial_balance)
+
+        return true
+
+      else
+        raise ArgumentError.new("You cannot create a money market account with less than a $#{ MINIMUM_BALANCE } initial deposit.")
+      end
+    end
```
katiespiders: Same comments here -- you can eliminate the if block by moving the `raise` to the top, and you don't need to return anything.

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+    attr_reader :transactions
+
+
+    def initialize(id, initial_balance)
+      if (initial_balance >= MINIMUM_BALANCE)
+        @transactions = 0
+
+        super(id, initial_balance)
+
+        return true
```
ktregina: no need for return

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      else
+        raise ArgumentError.new("You cannot create a money market account with less than a $#{ MINIMUM_BALANCE } initial deposit.")
```
ktregina: can be moved to top to eliminate the if/else block

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+      end
+    end
+
+
+    # add interest to account.
+    def add_interest(rate_percentage)
+      if validate_number(rate_percentage)
```
ktregina: can be called on its own, followed by the rest of the method (no if/else)

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+    # deposit money to account.
+    def deposit(amount)
+      unless balance_low? # if balance is low, don't do this stuff.
+        unless transaction_allowed? # if transaction is allowed, don't send a warning message!
+          warn "You cannot make any additional deposits this month. You have already reached the limit (#{ TRANSACTION_LIMIT })."
+
+          return @balance
+        else # when it is allowed, increment transactions counter.
+          @transactions += 1
+        end
```
ktregina, line 58: Unless/else blocks can be confusing to read (too many negatives). Reversing it to an if/else helps readability. 

```
if transaction_allowed?
  @transactions += 1
else
  warn "You cannot ... "
end
```

(`unless` without `else` is totally fine e.g. for cases like `unless balance_low?`, which reads better than `if not balance_low?`. It's just unless/else that you should avoid.)

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+      end
+
+      if super(amount) # call parent method to handle deposit.
+        return @balance #!Q does this need to be a conditional?
```
ktregina: I don't think so :)

ktregina:
(in case you've gotten email alerts about my comments and are checking -- I'm not done but will finish later! and you're doing great!)

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+        if balance_low?
+          apply_low_balance_fee
+        end
+
+        return @balance
+      else
+        return @balance
+      end
+    end
+
```
ktregina, line 95: I think that `@balance` can be moved to the end here, since it looks like every possible branch of the conditional will end with it returning `@balance`.

lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,141 @@
...
+
+    def balance_low?
+      if @balance < MINIMUM_BALANCE
+        return true
+      else
+        return false
+      end
+    end
```
ktregina, line 112: This can just return `@balance < MINIMUM_BALANCE`. That evaluates to true or false, so that's all you need to return. For me it took some getting used to writing such terse methods (having `@balance < MINIMUM_BALANCE` be the entire method just looks weird) but I think that is one of the awesome things about Ruby :)

ktregina:
Nice work (looks like really good test coverage too). I left a lot of comments, but it's really only a couple of things that I repeated a bunch -- mostly, when you write conditionals, think about whether you really need them. There's no technical problem with it, but complex conditionals can get confusing to read so it's best to get rid of or simplify them wherever you can. 

Let me know if you have any questions or if I can clarify anything!

lib/bank_accounts/money_market_account.rb
sojeri, line 112: I just wanted to say I've taken this to heart in future assignments -- even if I didn't respond to your advice in a timely fashion. I definitely consumed this advice and have been enjoying the one-line huh methods. Thank you! :)

lib/bank_accounts/savings_account.rb
sojeri, line 49: I was just rereading my updates to my code before merging the pull request, and I realized this comment still confuses me a bit. Can you give me some advice on when to use implicit returns? For my way of thinking, it usually makes more sense to spell out returns explicitly so I don't have to go back and research them later. Is it better to use implicit in this case -- when it involves checking out the code this code inherits from to make sure it's still returning the balance per the specifications?

sojeri:
Katie, sorry I didn't respond sooner! I really appreciate all the work you put into this review & have tried to implement all of it. I do have one question which I also touched on above re: when to use explicit vs. implicit returns.

lib/bank_accounts/savings_account.rb
ktregina, line 49: In this specific case, I actually would ignore the return value entirely. Generally, a method should be called to produce a "side effect" or to return a value, but not both. In this project, the `withdraw` method is called for its "side effect" of updating the `@balance` instance variable -- you don't want to also have to concern yourself with what it returns. 

I would write the parent method like so:

```
def withdraw(amount)
  return unless validate_withdrawal(amount)
  if (amount <= 0)
    warn "You cannot withdraw negative funds! That sounds like a transaction that should be a deposit."
    return
  end

  update_balance(-amount)
end
```

and the method in the subclass would just be:

```
def withdraw(amount)
  super(amount + WITHDRAWAL_FEE)
end
```

More generally: I think implicit vs explicit is a matter of style to a pretty significant degree, and probably not one that you need to be very worried about right now. There's nothing necessarily wrong with being more explicit than necessary, especially right now when you have so much new stuff being crammed into your brain, and you will probably get more comfortable with Ruby's love of implicitness with time and experience. 

As for the question about checking the code to make sure of what it's returning -- there is a [whole book](http://www.confidentruby.com/) on this. Put it on your to-read list if you're interested, but I would _not_ advise reading it now -- it's very advanced and very dense (I'm working my way through it over the course of months, with multiple re-readings of every chapter).

PR #8, opened 2015-05-31
Completed Bank Account Project

carjug:
I finished the Bank Account Project through all of the Breakfast requirements and most of Lunch (sans the reset_checks method). All of my specs are green. However I would like some feedback on how to better my specs in terms of length. I felt like I was repeating myself a bunch and I want to learn how to make specs that are more concise.

Thanks!


lib/classes/account.rb
```diff
@@ -0,0 +1,30 @@
...
+      if initial_balance < 0
+        raise ArgumentError.new, "Cannot have a negative account balance"
+      end
+      @id              = id
+      @current_balance = initial_balance
+    end
+
+    def withdraw(amount)
+      @current_balance -= amount
+      if amount > @current_balance
```
kariabancroft, line 15: Would you want to do the error check after you subtracted the amount or before?

lib/classes/account.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+      return @current_balance
+    end
+
+    def deposit(amount)
+      @current_balance += amount
+    end
+
+    def balance
+      @current_balance
```
kariabancroft, line 26: You could also name your instance variable `@balance` and then you wouldn't need this separate method, you could just use the attr_accessor

lib/classes/checking_account.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def withdraw_using_check(amount)
+      if amount > (@current_balance + 10)
+        raise ArgumentError.new, "You cannot overdraw more than -$10 of your account balance."
+      end
+      @check_use ||= 1
+       if @check_use <= 3
+        @current_balance -= amount
+      else @check_use > 3
```
kariabancroft, line 16: You can just an `else` here rather than specifying a condition

lib/classes/checking_account.rb
```diff
@@ -0,0 +1,26 @@
...
+    def withdraw(amount)
+      super
+      return (@current_balance -= 1)
+    end
+
+    def withdraw_using_check(amount)
+      if amount > (@current_balance + 10)
+        raise ArgumentError.new, "You cannot overdraw more than -$10 of your account balance."
+      end
+      @check_use ||= 1
```
kariabancroft, line 13: Nice use of the conditional assignment

lib/classes/savings_account.rb
```diff
@@ -0,0 +1,28 @@
...
+    def initialize(id, initial_balance)
+      super
+      if @current_balance < 10
+        raise ArgumentError.new, "Initial deposit must be greater than 10"
+      end
+    end
+
+    def withdraw(amount)
+      super
+      if amount > (@current_balance - 10)
```
kariabancroft, line 14: I think you will have a similar issue to what I pointed out in the base Account class because of the fact that you set the `@current_balance` before you check that it will not take the balance too low.

lib/classes/savings_account.rb
```diff
@@ -0,0 +1,28 @@
...
+        raise ArgumentError.new, "Savings cannot go below 10"
+      end
+      return (@current_balance -= 2)
+    end
+
+    def add_interest(rate)
```
kariabancroft, line 20: It seems like all of these variables are set as instance variables unnecessarily. I don't think `@rate` and `@interest` are needed anywhere else outside of this method.

spec/specs/account_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    it "checks if it has one parameter" do
+      expect(@account).to respond_to(:withdraw).with(1).arguments
+    end
+
+    it "does not allow the account to go negative" do
+      expect{@account.withdraw(15)}.to raise_error(ArgumentError)
+    end
+
+    it "returns :initial_balance" do
```
kariabancroft, line 35: I think you want this to be current balance, correct?

spec/specs/account_spec.rb
```diff
@@ -0,0 +1,66 @@
+require 'spec_helper'
+require 'bank_accounts'
+
+describe BankAccount::Account do
+
+  context "when .new is called" do
```
kariabancroft, line 6: These are good tests to verify the correct constructor setup

spec/specs/account_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      expect{@account.withdraw(15)}.to raise_error(ArgumentError)
+    end
+
+    it "returns :initial_balance" do
+      expect(@account.withdraw(4)).to eq(6)
+    end
+  end
+
+  context "when #deposit method is called"do
+    it "has a #withdraw method" do
```
kariabancroft, line 41: I think this is just a copy/pasta name mismatch

spec/specs/account_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    it "checks if it has one parameter" do
+      expect(@account).to respond_to(:deposit).with(1).arguments
+    end
+
+    it "returns :initial_balance" do
+      expect(@account.deposit(4)).to eq(14)
+    end
+  end
+
+  context "when #balance method is called" do
```
kariabancroft, line 54: It might be worthwhile to have another test that executes either the withdraw or deposit methods and then checks that the balance is correct

spec/specs/checking_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    it "incurs a fee of $2 when check_use is > 3" do
+      4.times do @checking.withdraw_using_check(1)
+      end
+      expect(@checking.balance).to eq(14)
+    end
+  end
+end
+
+
+
```
kariabancroft, line 52: You may still also want to include tests for the deposit method even though it is inheriting behavior from the overall Account.

lib/classes/account.rb
carjug, line 15: Probably would want to error check before :+1:

lib/classes/checking_account.rb
carjug, line 13: One of my fav things about this project ! :100:

lib/classes/savings_account.rb
carjug, line 14: Good point!

lib/classes/savings_account.rb
carjug, line 20: Haha! I think I was paranoid about when and when not to use them because I didn't understand the difference very well, so I just defaulted to using them always forever. 
I have a better idea now of when and when not to use an instance variable. :)

spec/specs/account_spec.rb
carjug, line 35: Yes. Lol.

spec/specs/account_spec.rb
carjug, line 6: I miss tests!

carjug:
Thanks for the feedback @kariabancroft <3
I look at this project now and kind of can't remember where I was at in my learning, so it's super difficult to imagine refactoring, but it's also awesome that I know what you're talking about in all of your comments :)

PR #9, opened 2015-05-31
Complete Bank Accounts Project - Lunch and part of Dinner

acmei:
I completed the requirements through Lunch and some Dinner. I added an additional method to round the balance values down to the hundredths position, this is instead of using the Float.round method which rounds up if the thousandths position is >= 5. I figured banks would round down. I'd like to learn how to DRY up the withdraw_using_check method in my CheckingAccount class. Also, I used pry to do some debugging.


roseaboveit:
I'm starting this review, it may take a few days but I'll post a top level message again when I'm complete.

Elffers:
@roseaboveit FYI, you can assign yourself to this PR in the right hand sidebar, rather than saying you're starting the review via comment :)

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,45 @@
...
+  class Account
+
+    attr_reader :id, :balance
+
+    # Rounds float value down to hundredths position
+    # For example 10.2599 rounds to 10.25
+    def round_to_hundredths(n)
+      n = (((n * 100).floor) / 100.0)
+      return n
+    end
```
roseaboveit, line 13: I like your round_to_hundredths implementation. You have clearly considered the context of your program.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,45 @@
...
+module BankAccounts
+  class Account
+
+    attr_reader :id, :balance
+
+    # Rounds float value down to hundredths position
+    # For example 10.2599 rounds to 10.25
+    def round_to_hundredths(n)
+      n = (((n * 100).floor) / 100.0)
+      return n
```
roseaboveit, line 12: Ruby has implicit returns so you don't need line 12 at all. 
The last line of the method is setting a variable so that variable is what will automatically be returned by the method.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      # Raises warning if withdraw amount is greater than present balance
+      if amount > @balance
+        raise ArgumentError.new "INSUFFICIENT FUNDS\nYour current balance is $#{round_to_hundredths(@balance)}."
+
+      # Return the updated balance once money is withdrawn from the account
+      else
+        @balance -= amount
+      end
+      return @balance = round_to_hundredths(@balance)
```
roseaboveit, line 35: Similarly to above, the implicit return rule exists here. So at the very least you can drop the word `return`

Another question to consider. At what points in the user flow do we need to round to hundredths? If the balance coming into this method already is rounded to hundredths and a manual withdrawal or deposit can't have more significant digits than pennies, do we need to round to hundredths at the end of a manual deposit or withdrawal?

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,45 @@
+require 'pry'
+
+module BankAccounts
+  class Account
+
+    attr_reader :id, :balance
```
roseaboveit, line 6: I like that you chose to consistently use attr_reader in your classes.

lib/bank_accounts/checking.rb
```diff
@@ -0,0 +1,55 @@
+require 'pry'
+
+module BankAccounts
+  class CheckingAccount < Account
+
+    attr_reader :balance, :checks_used
+
+    def initialize(id, initial_balance)
+      # inherit from Account class
+      super
```
roseaboveit, line 10: :+1: yay `super`

lib/bank_accounts/checking.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      # start with 0 checks used
+      @checks_used = 0
+    end
+
+    def withdraw(amount)
+      #incur a fee of $1 with each withdrawal
+      amount += 1
```
roseaboveit: Transaction fee in the instructions is $2. Always good to double check details before committing to master.

lib/bank_accounts/checking.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def initialize(id, initial_balance)
+      # inherit from Account class
+      super
+
+      # start with 0 checks used
+      @checks_used = 0
+    end
+
+    def withdraw(amount)
```
roseaboveit, line 16: Since there are multiple withdrawal methods it may be good to rename this one to be more specific. Is this withdrawal via ATM?

lib/bank_accounts/checking.rb
```diff
@@ -0,0 +1,55 @@
...
+      # no transaction fee for 3 or fewer checks
+      else
+        if (@balance - amount) < -10
+          raise ArgumentError.new "OVERDRAFT WARNING - only $10 overdraft allowed. Your current balance is #{round_to_hundredths(@balance)}."
+        else
+          @balance -= amount
+        end
+        return @balance = round_to_hundredths(@balance)
+      end
+    end
```
roseaboveit, line 36: Great job getting this working! That is the most important thing! 

You may notice that you have a lot of duplication in structure here. One option for refactor might be moving around the order of the control flow. Lines 31-36 are identical to lines 40-45. 

When I see these nestings I reach for a couple tools. 

One is reorganization. Here I'm thinking, do I really care about the number of checks used if they are overdrawn and we won't cash their check anyway? Perhaps we should check for that first and only deal with checks if we would process them. 

The other tool I reach for is abstraction. Can I pull out any chunks of code to make their own method to make this easier to read. One thing to note here is that in the regular withdrawal method we call super to reuse code that is already written in the parent class. Can we do that anywhere in the withdraw_using_check class?

lib/bank_accounts/savings.rb
```diff
@@ -0,0 +1,45 @@
...
+      # error raised if withdrawal causes balance to dip below $10
+      if (@balance - amount) < 10
+        raise ArgumentError.new "WARNING - You must have at least $10 in your savings account.\nYour current balance is $#{round_to_hundredths(@balance)}."
+      else
+        super
+      end
+      return @balance = round_to_hundredths(@balance)
+    end
+
+    def interest(rate)
```
roseaboveit, line 34: This method is the one that I think makes the most sense to apply the round_to_hundredths method

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    it "rounds initial balance floats > longer than 2 decimal spots, down to 2" do
+      account = BankAccounts::Account.new(0, 10.2599)
+      expect(account.balance).to eq(10.25)
+    end
+  end
+
+  context "When withdrawing money from bank account" do
+    let(:account) { BankAccounts::Account.new(0, 200) }
```
roseaboveit, line 22: Good job using a `let` block to dry up your test preparations.

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      expect(account.withdraw(36.2596)).to eq(113.74)
+    end
+
+    it "raises an error when the account is overdrawn" do
+      expect{ account.withdraw(500) }.to raise_error(ArgumentError, "INSUFFICIENT FUNDS\nYour current balance is $#{account.balance}.")
+    end
+  end
+
+  context "When depositing money into the bank account" do
+    let(:account) { BankAccounts::Account.new(0, 500) }
```
roseaboveit, line 36: Would it be possible to create an account object that could be placed at the `describe BankAccounts::Account` level instead of the context levels and cover all of the tests?

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    it "responds to deposit(amount)" do
+      expect(account).to respond_to(:deposit)
+    end
+
```
roseaboveit, line 41: Responding to #deposit is a good interim test, but one that is tested by later tests. Good for the Red & Green parts of TDD but can be removed in the refactor step once you have other tests that test this functionality. One way to think about it is thinking about when this test would fail and if there are other tests that would need to fail if this test failed. The next test would fail at the same time and would give a reasonable error message so this test is unnecessary in later stages.

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    it "returns an updated account balance when depositing money" do
+      expect(account.deposit(50)).to eq(550)
+      expect(account.deposit(150)).to eq(700)
+    end
+  end
+
+  context "When checking account balance" do
```
roseaboveit, line 48: "When checking account balance" is not really a context. Maybe you want a describe block for "checking account balance?

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    let(:account) { BankAccounts::Account.new(0, 10000) }
+
+    it "responds to balance" do
+      expect(account).to respond_to(:balance)
+    end
+
```
roseaboveit, line 54: Another candidate for test refactoring.

spec/lib/bank_accounts/checking_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      expect { checking.withdraw_using_check(150) }.to raise_error(ArgumentError, "OVERDRAFT WARNING - only $10 overdraft allowed. Your current balance is #{checking.balance}.")
+    end
+
+    it "adds a $2 fee when more than 3 checks are used" do
+      5.times { checking.withdraw_using_check(10) }
+      expect(checking.balance).to eq(46)
+    end
+  end
+
+  context "When a new month begins, checks used count is reset" do
```
roseaboveit: The context here is that a new month begins. The test is that the checks used count is reset.

spec/lib/bank_accounts/checking_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    let (:checking) { BankAccounts::CheckingAccount.new(0, 500) }
+
+    it "resets the number of used checks from 10 to 0" do
+      10.times { checking.withdraw_using_check(2.6899) }
+      expect(checking.checks_used).to eq(10)
+      expect(checking.reset_checks).to eq(0)
```
roseaboveit, line 49: Line 48 isn't strictly necessary. Unless you want to check that checks_used is functionality that isn't tested elsewhere but it is a different bit of functionality than the reset functionality so would be better tested above.

checking.reset_checks is actually part of the setup for this test so should probably be outside of the expect block.

Then you can set an expect for the checks_used method directly. While I think the result is the same, it will be easier to read and understand what is going on to have it be

10.times { checking.withdraw_using_check(2.6899) }
checking.reset_checks
expect(checking.checks).to eq(0)

spec/lib/bank_accounts/money_market_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    expect{ BankAccounts::MoneyMarketAccount.new(0, 10) }.to raise_error(ArgumentError, "MINIMUM BALANCE $10000 REQUIRED")
+  end
+ end
+
+ context "When withdrawing from a money market account" do
+  let (:money_market) { BankAccounts::MoneyMarketAccount.new(0, 20000) }
+
+  it "returns an updated account balance after withdrawing" do
+    expect(money_market.withdraw(200)).to eq(19800)
+    expect(money_market.withdraw(4800)).to eq(15000)
```
roseaboveit: What do you see as the benefit of running two expectations here?
If you are testing that the value persists, perhaps run the withdrawal as a setup and then have the expectation be about the balance.

roseaboveit:
Sorry it took so long to complete this code review.

You did an excellent job. First off congratulations on getting so far in the project while also keeping up with the testing. Your testing sense is quite good. You do a good job of knowing what needs to be tested and your process through your commits is quite good as well. 

I gave some feedback on the refactoring you asked about. Let me know if you have any additional questions about that.

You have made great progress in the proficiencies taught in this assignment. There was a lot to learn and you did some great implementation. 

I especially love that you did a lot of commits. You started especially strong with a solid discipline of TDD. Well done!

lib/bank_accounts/account.rb
acmei, line 12: Hi @roseaboveit! I was adding that just to be explicit. Is this something that other devs/programmers dislike?

acmei:
Thanks so much @roseaboveit for your thorough feedback. I really appreciate the time you took to look over my code and review it!

lib/bank_accounts/account.rb
roseaboveit, line 12: It probably varies developer to developer. I've noticed that in Seattle there is a tendency to not use more characters than you need. So using implicit returns save a whole line. Additionally for more complex methods, you can simply have the last line of the method be the variable that gets returned. The word `return` is extra unnecessary. It is a small thing. I was noticing other folks were discussing implicit returns and didn't want you to miss out on that feedback but I don't think it is hugely important either. Most importantly, when you are reading code, understanding implicit returns will be very helpful.

PR #11, opened 2015-05-31
Completed BankAccounts project (breakfast)

vikshab:
Completed breakfast of BankAccounts project.
Account class has 10 'green' specs
SavingsAccount class has 8 'green' specs 
I wrote detail comments for the code, hopefully it will be easy to read it
Module (bank_accounts.rb) has testing suits for both classes, so if you comment them out and run them in terminal it will display 'clear' picture of what the classes do.


lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,51 @@
...
+      # # if it's not then it raises ArgumentError
+      if id.class != Fixnum
+        raise ArgumentError
+      end
+
+      @id = id
+
+      # Checking if initial_balance is a positive number
+      # if it's not then it raises ArgumentError
+      if (initial_balance.class != Fixnum && initial_balance.class != Float) || initial_balance < 0
```
jnf, line 19: Most of the time, you should do all of your error checking at the same time; verify before you do any other work. I like that you are verifying the class of your input and raising an error, but don't do any work (assigning `@id` is work) until you have validated all your input.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,51 @@
...
+#_____________________________________________________________________________________________
+# Deposit method
+
+    def deposit(amount)
+
+      @balance += amount
+      return @balance
+    end
+
+    def balance
```
jnf, line 47: The `attr_accessor` on line 4 makes the method unnecessary. :)

lib/bank_accounts/savings_account.rb
```diff
@@ -0,0 +1,52 @@
...
+# Constant: fee $2
+      FEE = 2
+#__________________________________________________________________________
+# Initialize method
+
+    def initialize(id, initial_balance)
+      super(id, initial_balance)
+
+      # if balance is less than $10 it will raise ArgumentError
+      if @balance < 10
```
jnf, line 18: Same comment as above: this is a good validation, but you're doing it after you've already done work (the `super` call). If the balance is less than 10, there's no reason to run `super()`.

spec/account_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  describe "withdraw method" do
+   let(:account) {BankAccount::Account.new(5, 500)}
+
+    it "withdraws $300 and returns updated balance $200" do
+      expect(account.withdraw(300)).to eq 200
+    end
+
+    it "returns last updated balance if withdraw amount is bigger than balance" do
+      expect(account.withdraw(100)).to eq 400
+      expect(account.withdraw(600)).to eq 400
```
jnf, line 46: I really like this test. Good job. Verifying object state at each step of a process is very helpful.

jnf:
Good job on this project. Your specs are very well organized and easy to read. :) Let me know if you have any questions.

PR #12, opened 2015-05-31
Bank Account - Lunch Version

brittinator:
Bank Account - Lunch, Please run bank_account.rb from the root folder.  No gems required.

In the specs, I tried not to repeat myself, but I still made a number of bank account instances.  Writing the first few specs are hard because there's not code to talk about, but I tried to start simple and work up to more complex tests.  I used inheritance for similar attributes/methods as much as I could, but if there's some that I missed please let me know.


lib/bank_account/account.rb
```diff
@@ -0,0 +1,29 @@
+module BankAccount
+  class Account
+    attr_reader :balance
+
+    def initialize(id, initial_balance)
+      # cannot start a bank account with a negative balance
+      raise ArgumentError, "Invalid amount." if initial_balance < 0
```
jnf, line 7: Awesome. Good pattern: doing verification before doing work. Well done.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,29 @@
...
+      @balance = initial_balance
+      @account_id = id
+    end
+
+    def withdraw(amount)
```
jnf, line 12: You could DRY up this method a bit by leveraging a guard clause like...

``` ruby
def withdraw(amount)
  # the puts is nice, but not requried. In the future, we'd use the `warn` method
  if amount < @balance
    puts "..."
    return @balance 
  end

  # this expression returns the result of the operation,
  # and since it's the last line of the method, that result is the implicit return value
  @balance -= amount
end
```

lib/bank_account/account.rb
```diff
@@ -0,0 +1,29 @@
...
+        return @balance
+      else
+        @balance -= amount
+        puts "You balance is: $#{@balance}."
+        return @balance
+      end
+    end # withdraw
+
+    def deposit(amount)
+       return @balance += amount
```
jnf, line 25: This method is perfect.

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,46 @@
...
+    def withdraw(amount)
+
+      if amount > @balance
+        puts "Invalid. That amount is greater than your account balance."
+      else
+        super
+        new_balance = @balance - 1 # $1 transaction fee tacked on
+        @balance = new_balance
+      end
+      return @balance
```
jnf, line 19: Your use of `super` is spot on here, but I think you could do without the `new_balance` local variable. This is also a good candidate for a guard clause like in my comment above.

lib/bank_account/saving_account.rb
```diff
@@ -0,0 +1,30 @@
+module BankAccount
+  class SavingAccount < Account
+
+    def initialize(id, initial_balance)
```
jnf, line 4: The comment's great, but if there's no difference between the two, you don't even have to include `initialize` here. Inheritance will bring the base class's intializer into this class. :)

lib/bank_account/saving_account.rb
```diff
@@ -0,0 +1,30 @@
...
+      super
+    end
+
+    def withdraw(amount)
+
+      # savings account cannot dip below $10
+      if @balance - amount < 10
+        puts "Your balance cannot be less than $10 in your savings account."
+      else
+        super
```
jnf, line 15: Excellent user of `super` here. You identified what the varying behavior is (minimum balance requirement) and handled the logic well. Great job.

spec/lib/bank_account/account_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    let (:neg_account) {BankAccount::Account.new(20, -100)}
+
+    it '#self.new with a negative balance raises an ArgumentError' do
+      expect{neg_account}.to raise_error(ArgumentError, "Invalid amount.")
+    end
+  end # context
+
+  let(:account) {BankAccount::Account.new(20, 100)}
```
jnf, line 18: I almost missed this line when I was reading the spec. Generally, `let`s should be right at the top of the relevant `describe`/`context` block.

spec/lib/bank_account/account_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(account.deposit(20)).to eq(120)
+    end
+  end # deposit
+
+  context "#balance" do
+    it "returs current account balance of 100" do
+      expect(account.balance).to eq(100)
+    end
+  end
+
```
jnf, line 43: Since all the tests are against the same object (using the same `let`), you probably don't need to isolate them into `context` blocks. Doesn't break anything, of course, it's just unnecessary hierarchy.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,29 @@
+module BankAccount
+  class Account
+    attr_reader :balance
```
jnf, line 3: yay attr_reader!

spec/lib/bank_account/checking_account_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  end # withdraw_using_check
+
+  context "#reset_checks" do
+      let(:checker) {BankAccount::CheckingAccount.new(123, 5_000)}
+
+    it "resets the check counter back to zero" do
+      5.times do
+        checker.withdraw_using_check(200)
+      end
+      expect(checker.reset_checks).to eq(0)
```
jnf, line 61: I like this test! It'd be worth explaining with either a comment or another `expect` why the `5.times` iterator is relevant to the spec.

jnf:
Great job on this project. Most of my comments are just about organization and clarity; you handled the implementation and requirements really well. Have a look at my comments and let me know if you have any questions or concerns. :)

spec/lib/bank_account/account_spec.rb
brittinator, line 18: Thanks Jeremy, that's definitely good to know.

spec/lib/bank_account/account_spec.rb
brittinator, line 43: Also good to know.  I was wondering about how many levels to add for explanation.

spec/lib/bank_account/checking_account_spec.rb
brittinator, line 61: Oh yeah, I didn't even think about how this makes no sense by just looking at contexts and it blocks

lib/bank_account/account.rb
brittinator, line 12: Oh yeah, ordering helps in this case!

brittinator:
Thanks Jeremy, your comments are very helpful.  I'll try and bring forward the spec ordering into further projects, and look for squelchy code that may be done cleaner.

PR #14, opened 2015-06-01
Breakfast BankAccounts

ktrops:
Here's my breakfast version of BankAccounts.


lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,36 @@
...
+    if @balance - amount <= 0
+      puts "You cannot withdraw THAT much money."
+    else
+      @balance -= amount
+    end
+    return @balance
+    end
+
+    def withdraw_using_check(amount)
+      @checks += 1
```
kariabancroft, line 16: This is a lot of hard-code values, so this would be a great place to use constants

lib/bank_account/account.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+      return @balance
+    end
+
+    def deposit(amount)
+      @balance += amount
+      return @balance
+    end
+
+    def balance
```
kariabancroft, line 30: Since you're using an instance variable called `@balance` you could use an `attr_reader` for this variable and then you don't need a separate method

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,36 @@
+module BankAccount 
+  class CheckingAccount < Account
+    attr_accessor :checks
+
+    def withdraw(amount)
+    amount += 1
```
kariabancroft, line 6: Watch your indentation in this method

lib/bank_account/account.rb
```diff
@@ -0,0 +1,38 @@
+module BankAccount
+
+  class Account
+
+    def initialize(id, initial_balance)
+      @id = id
```
kariabancroft, line 6: Do you use this `@id` variable anywhere?

spec/lib/checking_account_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    it "knows about init from parent class" do
+      expect(BankAccount::CheckingAccount).to respond_to :new
+    end
+  end
+
+  context "withdraw methods methods" do
+    let(:checking_account) {BankAccount::CheckingAccount.new(6, 600)}
+
+    it "withdraws correct amount + $1" do
```
kariabancroft, line 15: Would use "fee" in your `it` descriptor rather than `$1` so that if the fee amount changed you wouldn't have to change your test

spec/lib/checking_account_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it "will not let your accout be <= 0" do
+      expect(checking_account.withdraw(599)).to eq 600
+    end
+
+    it "#withdraw_using_check" do
+      expect(checking_account.withdraw_using_check(50)).to eq 550
+    end
+
+    it "withdraw_using_check withdraws the fee after using 3 checks" do
+      3.times do |count|
```
kariabancroft, line 28: This is great! I like that it's checking the checks amount every time

spec/lib/checking_account_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+describe BankAccount::CheckingAccount do 
+  
+  context "CheckingAccount" do
+
+    it "knows about init from parent class" do
+      expect(BankAccount::CheckingAccount).to respond_to :new
+    end
+  end
+
+  context "withdraw methods methods" do
```
kariabancroft, line 12: Consider re-phrasing to just use "withdraw methods"

spec/lib/savings_account_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper.rb'
```
kariabancroft, line 1: Nice use of the `let` statement throughout all of your tests

spec/lib/savings_account_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect{(BankAccount::SavingsAccount.new(3, 5))}.to raise_exception(ArgumentError)
+    end
+  
+  end
+
+  context "balance" do
+
+    let(:savings_account) {BankAccount::SavingsAccount.new(4, 150)}
+
+    it "withdraws the correct amount + $2" do
```
kariabancroft, line 22: Would also recommend not using specific $ amounts in these tests in case the fee ever changes.

kariabancroft:
Nice job on this - your base Account class is well designed and that design propagates nicely throughout all of the subclasses. Tests are well-designed but I would recommend removing the use of specific fees in your tests in the case that the fee amounts would ever change.

lib/bank_account/account.rb
ktrops, line 30: Oh, that makes sense.

spec/lib/checking_account_spec.rb
ktrops, line 12: I think this was a typo?

ktrops:
Thanks @kariabancroft !

PR #15, opened 2015-06-01
Completed BankAccounts project to breakfast level

lindseyreno:
I completed the BankAccounts project to the breakfast level.  I started working on lunch and got through the first 5 bullet points.  Currently when I run rspec I have 22 examples and 1 failure which is the spec I wrote to work on the 6th bullet point of lunch.


lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,37 @@
+module BankAccounts
+
+  class Account
+    attr_accessor :id, :balance
+
+    def initialize(id, initial_balance)
+      # raises an error if initial_balance is negative
+      if initial_balance < 0
+        raise ArgumentError.new("Only positive initial balances allowed")
```
Elffers: :+1: Alternatively, you can also just say `raise ArgumentError, "Only positive ..."` instead of actually creating an instance of ArgumentError.

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def withdraw(amount)
+      # does not allow the account to go negative
+      # will output a warning message and return the original un-modified balance
+      if amount > @balance
+        puts "You can't withdraw what you don't have..."
+      else
+      # the input amount gets taken out of the account as result of an ATM transaction
+        @balance = @balance - amount
```
Elffers: Style tip: You can also say `@balance -= amount`, which does the same thing as line 23. Similarly, if you wanted to add something to a variable and reassign the new value to the same variable, you can do `foo += amount`

lib/bankaccounts/account.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+      # return value should be the updated account balance
+      return @balance
+    end
+
+    def deposit(amount)
+    # adds the input amount to the account balance as a result of an ATM transaction
+      @balance = @balance + amount
+    # return value should be the updated account balance
+      return @balance
```
Elffers, line 33: Ruby does implicit returns, so no need to actually say `return` here; just `@balance` will work.

lib/bankaccounts/checking_account.rb
```diff
@@ -0,0 +1,47 @@
+module BankAccounts
+
+  class CheckingAccount < Account
+      # attr_accessor :check_count
```
Elffers: This should be uncommented, otherwise you won't be able to read this instance variable or set it to a new value (such as in `reset_checks`).

spec/lib/checking_account_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
```
Elffers: Due to Rspec magic (see the first paragraph of your `spec_helper.rb` file), you shouldn't have to include this `require 'spec_helper'` line. Doesn't hurt, but just FYI :)

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'spec_helper'
```
Elffers: Due to Rspec magic (see the first paragraph of your `spec_helper.rb` file), you shouldn't have to include this `require 'spec_helper'` line. Doesn't hurt, but just FYI :)

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+describe BankAccounts::Account do
```
Elffers: If you want to avoid having to type out `BankAccounts::Account` everywhere, you can wrap the whole file in `module BankAccount`, the same way you did your classes, and everything should be namespaced properly. That is, 

```
module BankAccounts
   describe Account do
     ...
    end
end
```

lib/bankaccounts/checking_account.rb
```diff
@@ -0,0 +1,47 @@
...
+      # each withdrawal 'transaction' incurs a fee of $1 that is taken out of the balance.
+      # returns the updated account balance.
+      end
+      return @balance
+    end
+
+    def withdraw_using_check(amount)
+      # allows the account to go into overdraft up to -$10 but not any lower
+      # allowed three free check uses in one month, but any subsequent use adds a $2 transaction fee
+      fee = (@check_count <= 3) ? 0 : 2
```
Elffers, line 29: Nice use of the ternary operator! :clap:

lib/bankaccounts/savings_account.rb
```diff
@@ -0,0 +1,39 @@
+module BankAccounts
+
+  class SavingsAccount < Account # inherits from Account class
+
+    def initialize(id, initial_balance)
+      super
+      # raises an error if initial_balance is less than 10
+      if initial_balance < 10
+        raise ArgumentError.new("Initial balances must be at least 10")
+      end
```
Elffers, line 10: Good logic :+1: Suggestion: Have you thought about extracting this into a separate method that can be called from `initialize`?

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  it "returns updated balance of 120 with deposit of 20" do
+    expect(@account.deposit(20)).to eq(120)
+  end
+
+  it "returns current balance of 100" do
+    expect(@account.balance).to eq(100)
+  end
+
+end
```
Elffers, line 40: Generally, I would recommend organizing your tests per method, which you can group using `context` or `describe` blocks. So for example, you'd have `describe "#withdraw" do` with all tests within that scope that would call the `#withdraw` method. The description in the `it` block should also not be specific to the details of your test data. So rather than saying:

```
it "returns 70 when withdrawing 30" do
```

you might say something more generic like 

```
describe '#withdraw' do
  it "deducts the given amount from the account balance" do
    account.withdraw 10
    expect(account.balance).to eq 70
  end
end
```

That way, if you ever decide to change the numbers or add more tests, the description is still applicable.

Basically, you want your test's description to match your test's assertion.

spec/lib/checking_account_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    expect(@checking_account.withdraw_using_check(115)).to eq(100)
+  end
+
+  it "returns -5 when when withdrawing 105 by check" do
+    expect(@checking_account.withdraw_using_check(105)).to eq(-5)
+  end
+
+  #The user is allowed three free check uses in one month, but any subsequent use adds a $2 transaction fee
+  it "returns 58 after 4 #withdraw_using_check(10)" do
+      expect(4.times {@checking_account.withdraw_using_check(10)}).to eq(58)
```
Elffers: I would have the setup evaluated before you set up your expectation, so that it is clear what you are trying to assert in your test. e.g., 

```
it "deducts a transaction fee after max number of free checks are used" do
  4.times do
     account.withdraw_using_check 10
  end
  expect(acccount.balance).to eq 58
end
```

spec/lib/checking_account_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+describe BankAccounts::CheckingAccount do
+
+  it "responds to #new" do
+    expect(BankAccounts::CheckingAccount).to respond_to :new
+  end
+
+  before :each do
+    @checking_account = BankAccounts::CheckingAccount.new("Name", 100)
+  end
```
Elffers: Rather than use an instance variable, you can use a `let` block to access objects set to variables. E.g., 

```
let(:checking_account) { BankAccounts::CheckingAccount.new("Name", 100) }
```

This allows you to access the variable `checking_account` in any test within the scope of your describe block.

lib/bankaccounts/checking_account.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def withdraw(amount)
+      # does not allow the account to go negative.
+      # will output a warning message and return the original un-modified balance
+      if amount > @balance
+        puts "You can't withdraw what you don't have..."
+      else
+      # the input amount gets taken out of the account as result of an ATM transaction.
+        @balance = @balance - (amount + 1)
```
Elffers: You may consider extracting the idea of a transaction fee, so that your code is more readable.

PR #16, opened 2015-06-01
Khn/master

knguyen0710:
- This submission is for the dinner tier.
- It is to be run from the base file bank_account.rb

I feel unsure about my specs, so I'd love a little additional feedback on them.

For MoneyMarket, I was able to use a lot of similar methods from other classes to either incorporate "super" or condense my code with a better understanding of how to approach the problem. The "withdraw" method uses a lot of other methods but is considerably shorter than some of my previous methods.


lib/bank_account/account.rb
```diff
@@ -0,0 +1,26 @@
+module Bank
+  class Account
+    attr_accessor :id, :balance
+
+    def initialize(id, initial_balance)
+      # raises error for negative initial balance
+      raise ArgumentError.new "Invalid initial balance" unless initial_balance > 0
+      @id = id
```
khjrtbrg, line 8: It might be nice to isolate the error raising into its own method that either returns `initial_balance` or raises the error. For instance, maybe something like this:

``` ruby
  @id = id
  @balance = set_initial_balance(initial_balance, 0)
end

def set_initial_balance(balance, min_amount)
  if balance > min_amount
    balance
  else
    raise ArgumentError.new "Invalid initial balance"
  end
end
```

Having a method like `set_initial_balance` might come in handy in your other classes where you're repeating similar behavior.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,26 @@
...
+      @balance -= amount
+      # for withdrawal greater than balance
+      # gives warning error and returns original balance
+      if @balance < 0
+        puts "Insufficient funds. Your remaining balance is:"
+        return @balance += amount
+      else
+        return @balance
+      end
+    end
```
khjrtbrg, line 25: This totally works, but you might want to think about checking if there's enough remaining in `@balance` before subtracting, just in case anything goes haywire before you can add the withdrawn amount back in.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  end
+end
```
khjrtbrg, line 28: I'm not seeing a `#deposit(amount)` method in here (it's in the Breakfast level for this class)?

spec/lib/bank_account/account_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  let(:account) {Bank::Account.new("check_it", 1000)}
+
+  describe "attributes" do
+    it "responds to :balance" do
+      expect(account).to respond_to :balance
+    end
+
+    it "responds to :id" do
+      expect(account).to respond_to :id
+    end
```
khjrtbrg, line 13: This test and the one before it are great tests to start with, but you don't really need them. Your tests on lines 15 and 19 implicitly cover these tests.

spec/lib/bank_account/account_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(account.id).to eq("check_it")
+    end
+
+    it "has an initial balance of 1000" do
+      expect(account.balance).to eq(1000)
+    end
+
+    it "returns 750 for a 250 withdrawal" do
+      expect(account.withdraw(250)).to eq(750)
+    end
```
khjrtbrg, line 25: For this test and the ones on lines 15 and 17, I'd recommend making your it statement a little more descriptive of the general behavior you're testing rather than what you expect from your test data. For instance, instead of:

``` ruby
it "returns 750 for a 250 withdrawal" do
```

you could do:

``` ruby
it "withdraws the specified amount" do
```

It makes it easier for someone skimming your tests to see what kinds of behaviors they can expect from your class. It'll also be helpful for you to remind yourself what the behavior should be as your projects get bigger. Also, more general, behavior-based statements mean you'll have to do less clean up if you ever decide to change your test data.

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,28 @@
...
+        return @balance
+      end
+    end
+
+    # calculates interest on balance where rate is in %
+    def add_interest(rate)
+      @accrued_interest = @balance * (rate/100.to_f)
+      @balance += @accrued_interest
+      return @accrued_interest
+    end
```
khjrtbrg, line 25: I'm not convinced you need to make `accrued_interest` an instance variable (the `@` turns it into one), since you aren't using it outside of this method and you reset it every time this method is called. Also Ruby should handle the conversion to a float if it's needed on line 22, so you don't really need to do the extra conversion at the end there.

spec/lib/bank_account/savings_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'bank_account'
```
khjrtbrg: This is nit-picky, but this file's name should follow the same naming convention as the file you're testing (i.e. `savings_account_spec.rb` for `savings_account.rb`)

spec/lib/bank_account/savings_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+describe Bank::SavingsAccount do
+  context "attributes" do
```
khjrtbrg, line 4: It looks like all the tests in this file are wrapped in this `context` block. You could probably get rid of it, or just wrap the ones related to attributes in it. I tend to use `context` blocks for setting up edge case situations, rather than for testing the happy path situation.

spec/lib/bank_account/savings_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    it "has an initial balance of 100" do
+      expect(savings.balance).to eq(100)
+    end
+
+    it "returns 48 for a withdrawal of 50" do
+      expect(savings.withdraw(50)).to eq(48)
+    end
+
+    context "accruing interest" do
+      let(:more_savings) {savings.add_interest(10)}
```
khjrtbrg, line 28: If you only have one test that uses this `let` statement, you could just make a local variable inside of your test.

spec/lib/bank_account/savings_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    context "low balance errors" do
+      it "raises 'Insufficient Funds' for balance < 10" do
+        expect{Bank::SavingsAccount.new("so_broke", -10)}.to raise_error(ArgumentError)
+      end
+
+      it "returns original balance of 100 for overdraft" do
+        expect(savings.withdraw(1000)).to eq(100)
+      end
+    end
```
khjrtbrg, line 45: While these are similar behaviors, you might want to move these into separate contexts. When you dig into them a bit, they're actually testing different behaviors: one is that an error is raised when needed as you initialize the object, whereas the other is that it doesn't overdraft when withdrawing. Grouping the first with the other tests dealing with initialize behavior and the second with withdrawing tests would also make the tests slightly easier to follow.

spec/lib/bank_account/savings_spec.rb
khjrtbrg, line 28: You also might not need a separate context for this. Has something super wild happened in your setup for this test that differs from the other tests?

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,28 @@
...
+    # charges $2 for every withdrawal
+    def withdraw(amount)
+      @balance -= (amount + 2)
+      if @balance < 10
+        puts "Savings can not fall below $10.Your remaining savings amount is: "
+        return @balance += (amount + 2)
+      else
+        return @balance
+      end
+    end
```
khjrtbrg, line 18: This works, but just as a FYI (and I'm not necessarily advocating this), this method could have been written like this to avoid duplicating very similar code in your `Account` class:

``` ruby
def withdraw(amount)
  amount += 2
  super
end
```

spec/lib/bank_account/checkings.rb
```diff
@@ -0,0 +1,37 @@
+require 'bank_account'
```
khjrtbrg, line 1: @knguyen0710 It looks like this mostly the same as `checking_spec.rb`, so I'm going to look at that one instead. If you aren't using this file, you might want to remove it!

spec/lib/bank_account/checking_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect{Bank::CheckingAccount.new("broke_joke", -10)}.to raise_error(ArgumentError)
+      end
+
+      it "returns balance of 100 for ATM overdraft of 1000" do
+        expect(checkings.withdraw(1000)).to eq(100)
+      end
+
+      it "returns balance of 100 for check withdrawal of 500" do
+        expect(checkings.withdraw_using_check(500)).to eq(100)
+      end
```
khjrtbrg, line 39: It would be great to have more tests for this class, since you have a lot more edge cases going on. For instance:
- When a user make a check withdrawal within their overdraft limit
- When `time_passes`
- When the user has already used up their three checks for their month

lib/bank_account/checking.rb
```diff
@@ -0,0 +1,68 @@
...
+    # after which it is reset to the first of the month
+    # and number of checks used is reset to 0
+    def time_passes
+      if @day_of_month == 30
+        @day_of_month = 1
+        reset_checks
+      else
+        @day_of_month += 1
+      end
+    end
```
khjrtbrg, line 65: You might want to revisit how you're handling the 3 free check withdrawals every month constraint. With your existing code, what happens if a user doesn't make a withdrawal every day? Or what happens for months that aren't 30 days long? Is there another way you could keep track of when checks are used?

lib/bank_account/money_market.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+    end
+
+    # limits transactions to 6
+    # otherwise, adds to transaction counter
+    def transaction
+      if @transaction_count > 6
+        return "NO MORE TRANSACTIONS"
+      else @transaction_count += 1
+      end
```
khjrtbrg, line 68: I think your users would be able to get 7 transactions a month if it's written like this!

lib/bank_account/money_market.rb
```diff
@@ -0,0 +1,87 @@
...
+      else
+        transaction
+        return @balance += amount
+      end
+    end
+
+    # inherits interest method from SavingsAccount
+    def add_interest(rate)
+      time_passes
+      super
```
khjrtbrg, line 52: :+1: Nice! When you're dealing with bigger projects, this pattern might make your classes tightly coupled and a little trickier to work with as you add features. It works great for this project, but just something to be aware of!

lib/bank_account/money_market.rb
```diff
@@ -0,0 +1,87 @@
...
+      time_passes
+      # allows deposit transaction for balance below lower limit
+      # without adding to count
+      if @balance < 10000
+        return @balance += amount
+      # adds to transaction if balance is above lower limit
+      else
+        transaction
+        return @balance += amount
+      end
```
khjrtbrg, line 46: It looks like you write `return @balance += amount` twice. You might want to try rewriting the conditional so you only need it once.

spec/lib/bank_account/money_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      it "returns 'ACCOUNT FROZEN' for balance below 10000" do
+        expect(market.withdraw(25000)).to eq("ACCOUNT FROZEN: Must make a deposit.")
+        expect(market.balance).to eq(4900)
+      end
+
+    end
+  end
+end
```
khjrtbrg, line 31: I think you could use more tests for your edge cases. Try thinking of examples that might break your code for each requirement. If you phrase your `it` statements as behaviors, it might be easier to see where you could use more tests. Everywhere you have a conditional in your code, you should try to write tests covering each possible outcome.

Also, when you're using inheritance, you can get away with not retesting things tested in the base class, unless you change functionality.

khjrtbrg:
Nice work, @knguyen0710! :tada:

It looks like you're getting comfortable with inheritance, especially in your `MoneyMarketAccount` class. You could potentially leverage inheritance even more for some of the more repetitive methods (like `withdraw` and raising errors for insufficient initial balances).

It looks like you could use more tests for the lunch and dinner requirements that were more edge case-y. When writing tests, try to think of all the possible paths through your method and try to write tests to cover each of those paths.

It looks like the filenames for your specs didn't always mimic the filenames of the files you were testing. Being consistent with that will be super helpful as your projects get bigger!

The only other big thing I noticed was that you might want to revisit how you're handling monthly limits. `time_passes` works, but only if you can guarantee that it will be run exactly one time a day and all your months have the same number of days. Another way to approach the constraint is to keep track of when transactions happen with DateTime objects and then check if there are more transactions in the previous month than permitted.

I'm not sure if nesting your classes in a module was a requirement, but it was a nice organizing touch!

Good job!

lib/bank_account/account.rb
knguyen0710, line 25: And I realized this also made the code a little longer than it need to be too, so I'll definitely change that.

lib/bank_account/account.rb
knguyen0710, line 28: Yikes. I'll be sure to add it.

spec/lib/bank_account/account_spec.rb
knguyen0710, line 13: Which test is better form to write? The "responds to" or the name of an example account?

lib/bank_account/savings_account.rb
knguyen0710, line 25: I remember having trouble getting the calculation to work because it kept returning 0 so I added the float conversion and that seemed to help. After that, I didn't touch it again because it was working.

spec/lib/bank_account/account_spec.rb
khjrtbrg, line 13: I'd say the name of the example account, since that covers both that it has an attribute called `name` and that it's returning what you expect it to return.

PR #17, opened 2015-06-01
Bdl/master

Brennaleker:
I completed the project requirements through dinner.

Thank you for taking the time to review and comment on my code.

-Brenna


lib/bank_account/account.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def initialize(id, initial_balance)
+      ## optimize raise_error, initialize
+      if initial_balance < 0
+        raise ArgumentError.new("You cannot open an account with a negative balance.")
+      else
+      @id = id.round(2)
+      @initial_balance = initial_balance.round(2)
+      @balance = initial_balance.round(2)
+      end
```
kariabancroft, line 12: Since you're using a `@balance` instance variable you can actually just replace your `initial_balance` attr_accessor with an attr_accessor for `balance` and then you only need the local `initial_balance` var in the initialize.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,44 @@
...
+        puts "Nonsufficient funds."
+        balance
+      else
+      @balance -= amount
+      balance
+      end
+    end
+
+    def deposit(amount)
+      amount = amount.round(2)
```
kariabancroft: Since you're using `.round(2)` in several places you might want to create a method that handles this rounding instead. Then, if you needed to change the number of decimal places your system supported you would only need to change it in one place rather than multiple.

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,44 @@
+module BankAccount
+  class CheckingAccount < Account
+  attr_accessor :check_count, :balance, :id
+
+    def initialize(id, initial_balance)
+      if initial_balance < 0
+        raise ArgumentError.new("You cannot open an account with a negative balance.")
+      else
+      super
```
kariabancroft, line 6: This is a nice use of the base classes functionality - but you could remove lines 6-8 because the account classes initialize method already has that functionality.

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,44 @@
...
+      @check_count = 3
+      end
+    end
+
```
kariabancroft, line 9: Watch your indentation in the two methods below - you want to keep the indentation in between the `else` and `end` consistent with the indentation you see between the `if` and `end`.

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,44 @@
...
+      else if check_count > 0
+      @balance -= amount
+      @check_count -= 1
+      balance
+      else
+      @balance -= (2 + amount)
+      balance
+        end
+      end
+    end
```
kariabancroft, line 32: This method is very difficult to read because of the combination of the use of the `@balance` variable and `balance` method usage. Would recommend refactoring this to use only one variable/method with that name.

lib/bank_account/money_market_account.rb
```diff
@@ -0,0 +1,66 @@
+module BankAccount
+  class MoneyMarketAccount < Account
+    attr_accessor :id, :initial_balance, :transaction_count
+
+    def initialize(id, initial_balance)
+      if initial_balance < 10_000
+        raise ArgumentError.new("You are too poor, get out of here peasant!")
```
kariabancroft, line 7: LOL

spec/account_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect{BankAccount::Account.new(1,-5)}.to raise_error(ArgumentError)
+    end
+  end
+end
+
+describe "instance methods" do
+  let(:account) { BankAccount::Account.new(1,250)}
+
+  context "withdraw method removes money from account and returns updated balance" do
+    it "returns an updated balance of 45.00" do
```
kariabancroft, line 23: Since your test itself is what is checking for a specific balance, I would be less inclined to include that in your `it` descriptor.

spec/account_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+
+  context "deposit method adds money to account and returns updated balance" do
+    it "returns the updated balance of 500" do
+      expect(account.deposit(250)).to eq(500)
+    end
+  end
+
+  context "bank cannot process deposits of more than $2 million" do
```
kariabancroft, line 40: I like this test and this check that you added in!

spec/checking_account_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'spec_helper'
+
+describe "checking account attributes" do
```
kariabancroft, line 3: I'd add a test for the `balance` for each of your account tests

spec/money_market_account_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    it "allows a user with 0 transactions left to bring their account into good standing ($10k)" do
+        6.times do money_market_account.withdraw(2_000)
+        end
+      expect(money_market_account.deposit(3_000)).to eq(10_900)
+    end
+  end
+
+  context "allowable transactions are reset monthly" do
+    it "resets available transactions to 6" do
+      money_market_account.reset_transactions
```
kariabancroft, line 72: It would make sense to check the transaction count before the `reset_transactions` method is called to verify that the value has actually changed

kariabancroft:
Nice job on this! There is some confusing code with relation to the `@balance` instance variable and the `balance` method you created so I would strongly recommend trying to refactor that to use only one or the other. Your tests cover a lot of good cases so it looks good.

PR #18, opened 2015-06-01
BankAccounts (Lunch)

QuartzBrandi:
This is my BankAccount assignment. I completed up to the Lunch tier. I'm still having trouble working with rspec. (I have a few commented-out questions scattered throughout my rspec.) I kept all my rspec tests (instead of cleaning up old ones) in favor of seeing my process and my work.


lib/bank_accounts.rb
```diff
@@ -0,0 +1,11 @@
+module BankAccounts
+
+  # NOTE TO SELF:
+  # require_relative is AMAZING -- doesn't matter what folder you're in terminal
+  # as long as you reference this file correctly, these will work
+  # "require" is dependant on terminal location
+  require_relative './bank_accounts/account'
+  require_relative './bank_accounts/savings_account'
+  require_relative './bank_accounts/checking_account'
+
```
Elffers, line 10: Since rspec automatically does the proper file loading here, you should even be able to simply say `require "bank_accounts/account"` etc. :)

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'spec_helper'
+
```
Elffers, line 2: Rspec's magic also makes it such that you shouldn't have to `require 'spec_helper'` here--it show know which tests to run.

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+describe BankAccounts::Account do
+
```
Elffers, line 4: Note: If you wanted to avoid having to say `BankAccounts::Account` every time in your tests, you can wrap the entire file in `module BankAccounts` (the same way you wrapped the class you're testing) and everything should be properly namespaced.

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  it "responds to #new" do
+    expect(BankAccounts::Account).to respond_to(:new)
+  end
+
+  it "responds to #new with two arguments" do
+    expect(BankAccounts::Account).to respond_to(:new).with(2).arguments
+  end
+
+  let(:bob) { BankAccounts::Account.new(1, 1000) }
```
Elffers, line 13: Ha--Is there a reason you named this variable `bob`? Generally you'd probably want to use something that more accurately describes the object, so as boring as it may be, you might be better off just using `account` ;)

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    expect(bob.id).to eq(1)
+  end
+
+  it "contains @balance" do
+    expect(bob.balance).to eq(1000)
+  end
+
+  it "cannot initialize with a negative initial_balance" do
+    expect{BankAccounts::Account.new(2, -1000)}.to raise_error(ArgumentError)   #raise_error is the same as raise_exception
+  end   # Why does raise_error use expect{} vs expect()???
```
Elffers, line 25: The use of curly braces `{ }` as the argument to `expect` means that you're expecting whatever gets executed within those braces to do a thing (in this case, raise an error), whereas `expect` with parens `()` means that Rspec is going to compare the _return value_ of whatever is inside the parens with whatever you are setting up your expectation to be.

Here's an answer on [stackoverflow](http://stackoverflow.com/questions/19960831/rspec-expect-vs-expect-with-block-whats-the-difference) that might be a bit easier to understand than the rspec docs. Let me know if that makes sense!

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  it "#withdraw subtracts 'amount' from the balance & returns balance" do
+    expect(bob.withdraw(100)).to eq(900)
+  end
+
+  context "#withdraw cannot subtract more than balance:" do
+
+    it "outputs a warning message" do
+      expect{bob.withdraw(1100)}.to output("Cannot withdraw amount; not enough money in account.\n").to_stdout
+    end   # NOTE TO SELF: output().to_stdout checks the method bob.error is printing the string
```
Elffers, line 39: Interesting, I didn't know about `output` :) Another way you can test the actual error message you want to show when the specific error is raised is by saying simply:
`expect { bob.withdraw(1100) }.to raise_error ArgumentError, "Cannot withdraw amount; not enough money..."`. See matching error messages in the [docs](https://www.relishapp.com/rspec/rspec-expectations/v/3-3/docs/built-in-matchers/raise-error-matcher#matching-message-with-`with-message`).

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    it "& returns original balance" do
+      expect(bob.withdraw(1100)).to eq(1000)
+    end   # 1000 is original balance
```
Elffers, line 43: Based on your comment, one thing you might consider doing is that since you use `1000` so much, you could assign that to a variable in a `let` block (i.e. `let(:original_balance) { 1000 }`). That way, you can just refer to `original_balance` in any test that uses it as a comparison, and makes your test more descriptive. The other added bonus is that if you ever decide that 1000 isn't a great amount to use for whatever reason, you can just change the value in your `let` block and not have to change it everywhere it appears in your tests.

spec/lib/bank_accounts/account_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  it "#balance returns current account balance" do
+    expect(bob.balance).to eq(1000)
+  end
+
+  it "#balance returns current account balance (after manipulation of balance)" do
+    account = BankAccounts::Account.new(3, 500)
+    account.withdraw(400)
+    account.deposit(100)
+    expect(account.balance).to eq(200)
+  end   # QUESTION: Is there a way to execute multiple methods in a row on the same object without typing them all out?
```
Elffers, line 72: This brings up a good point. Generally, a good rule of thumb to abide by when organizing your tests is to have a `describe` block for each method. So, you'd have `describe '#withdraw' do` with all your tests that would call `withdraw` grouped under there, and so forth. It's probably not a ideal to call two different methods within the same test, bc then it might be unclear what behavior you're testing.

spec/lib/bank_accounts/checking_account_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      expect(tom.withdraw_using_check(1011)).to eq(1000)
+    end
+
+  end
+
+  context "#reset_checks changes the @checks_used to 0:" do
+    # Had a really hard time with this rspec.
+    # Why wouldn't the 'expect' on line 65 work
+    # if I put lines 60-64 outside the it-do block?
+
```
Elffers, line 58: It wouldn't work because it's outside the scope of the test. `it` is technically a method provided by Rspec, in the same way you have methods on classes. So any state you set up for your expectation needs to be accessible from the scope of that expectation.

If you wanted to reuse a set up/state, you can put lines 61-63 in a `before` block within the context you are testing. That is, on line 55:

```
let(:account) { BankAccounts::CheckingAccount.new(3, 1000) }
before do
  3.times do
      account.withdraw_using_check(100)
  end
end
```

However, since you only need this state for one test, it's probably not necessary here.

spec/lib/bank_accounts/checking_account_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        account.withdraw_using_check(100)
+      end
+      account.reset_checks
+      expect(account.checks_used).to eq(0)
+    end
+
+  end
+
+  # Is it usual to copy and paste rspecs from a parent class?
+
```
Elffers, line 79: I've often heard that the DRYness principle doesn't apply to tests, but if you really wanted to be DRY, you could look up shared_examples in the Rspec docs.

lib/bank_accounts/savings_account.rb
```diff
@@ -0,0 +1,21 @@
+module BankAccounts
+
+  class SavingsAccount < Account
+
+    # ADDS INTEREST TO BALANCE & RETURNS INTEREST AMOUNT
+    def add_interest(rate)
+      interest = @balance * (rate/100)
+      @balance += interest
+      return interest
```
Elffers, line 9: Since Ruby has an implicit return of the last line of the method, you don't need to say `return`.

lib/bank_accounts/savings_account.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    private
+
+    def bank_info
+      @min_balance = 10
+      @atm_withdraw_fee = 2
+    end
+
```
Elffers: Where do you call `bank_info`? And what does it return?

You might consider using Constants instead of instance variables here. Since things like the `min_balance` and `atm_withdraw_fee` are things that don't change (or reflect state) for an instance of a `SavingsAccount`, setting it as a top-level constant might be a better option.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,58 @@
...
+      raise ArgumentError, "below allowed minimum" if num < @min_balance
+    end
+
+    # VARIABLES CREATED DURING INITIALIZATION
+    # SEPERATED FROM INITIALIZE SO CAN CHANGE WHEN INHERITED
+    def bank_info
+      @min_balance = 0
+      @atm_withdraw_fee = 0
+    end
+
```
Elffers, line 55: Good instinct in making this a private method. However, you might consider using constants (in the classes that inherit from this one, e.g.) instead of instance variables here. Since things like the `min_balance` and `atm_withdraw_fee` are things that don't change (or reflect state) for an instance of a bank account, setting it as a top-level constant might be a better option.

lib/bank_accounts/account.rb
```diff
@@ -0,0 +1,58 @@
...
+    # ADD AMOUNT TO BALANCE & RETURN BALANCE
+    def deposit(amount)
+      neg_num_guard(amount)
+
+      @balance += amount
+    end
+
+    private
+
```
Elffers, line 38: Good use of private methods :+1:

Elffers:
Overall, great job! One thing to consider: How would you reset checks based on what month it is? Can you find a gem that deals with testing methods that are time-sensitive?

spec/lib/bank_accounts/account_spec.rb
QuartzBrandi, line 25: Yeah! It does, thanks for explaining that!

spec/lib/bank_accounts/account_spec.rb
QuartzBrandi, line 43: Good to know!

QuartzBrandi:
Thanks for the info on how to handle inherited constants! I had a lot of trouble with that.

Maybe use chronic for time-sensitive testing? I haven't used it yet, so I'm not sure.

PR #19, opened 2015-06-01
[Lunch] BankAccounts

catchingash:
This completes the lunch requirement of the BankAccounts project. All tests pass.

requires 'pathname'

I'd really love feedback on these things in particular:
  I added additional parameters with default values for many of the methods so that the child classes could use these methods, rather than overwriting them. Is that advisable in this situation?
  On a related note, I played with different ways of overwriting the base class method's parameters [see: lib/bankaccount/savingsaccount.rb, lines 9-18]  - any insight into which way is preferred would be great. The option on line 9 seems prettier, but it seems like it could be risky to open that up to being accidentally passed a parameter and overwriting the values. The option on line 17 avoids this issue.


lib/bankaccount.rb
```diff
@@ -0,0 +1,9 @@
+module BankAccount
+  require 'pathname'
+
+  ROOT_DIR = Pathname.new(__dir__).parent
+  require ROOT_DIR.join('lib/bankaccount/account')
+  require ROOT_DIR.join('lib/bankaccount/savingsaccount')
+  require ROOT_DIR.join('lib/bankaccount/checkingaccount')
+
+end
```
Elffers, line 9: Using Pathname is pretty heavyweight here. Since rspec already does a lot of the proper file loading for you, you should just be able to `require 'bankaccount/account'` etc.

lib/bankaccount/account.rb
```diff
@@ -0,0 +1,31 @@
+module BankAccount
+  class Account
+    attr_reader :id, :balance, :initial_balance
+
+    def initialize(id, initial_balance, required_opening_balance = 0)
+      # account cannot be created with negative balance [breakfast]
```
Elffers, line 6: Using default arguments is one way to handle not having to override the same method in classes that inherit from a parent class. However, in this case, since the only type of account that has a required opening balance is `SavingsAccount`, making it a concern of the `Account` class might be considered bad practice, since it goes against OO principles. I might suggest using a constant in SavingsAccount that the `initialize` method can check there before calling `super` into the parent method.

lib/bankaccount/savingsaccount.rb
```diff
@@ -0,0 +1,29 @@
...
+    def withdraw(amount, transaction_fee = 2, min_balance = 10) # [breakfast]
+      # each withdrawal incurs a $2 fee (taken out of balance) [breakfast]
+      super
+    end
+
+    # def withdraw(amount) # [breakfast]
+    #   # each withdrawal incurs a $2 fee (taken out of balance) [breakfast]
+    #   super(amount, 2, 10) # (amount, transaction_fee, min_balance)
+    # end
+
```
Elffers, line 19: See comment on `lib/bankaccount/account.rb:6`. I would suggest using constants that only the child class knows about so that you don't have to use default arguments. That way you can inherit methods from the parent class more cleanly instead of having to overwrite them or put in custom arguments. This will make it particularly nice bc then you won't be able to accidentally put in the wrong amount for the argument `transaction_fee`  when you call `withdraw`.

lib/bankaccount/checkingaccount.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def check_fee
+      # 3 free checks per month; $2 transaction fee after that [lunch]
+      if @checks_used < 3
+        transaction_fee = 0
+      else
+        transaction_fee = 2
+      end
+    end
```
Elffers, line 33: Style tip: You can DRY up your code here by saying something like:

```
transaction_fee =
if @checks_used < 3
  0
else
  2
end
```

You might also consider using the [ternary operator](http://ruby-doc.org/core-2.2.0/doc/syntax/control_expressions_rdoc.html#label-Ternary+if) in Ruby, though some find it harder to read:
`@checks_used <  3 ? 0 : 2`

This would eliminate the need to assign the return value to a local variable at all.

spec/lib/account_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        expect(@my_account.initial_balance).to eq(125)
+      end
+    end
+
+    context 'when a negative account balance is passed in' do
+      it 'raises an ArgumentError' do
+        expect { BankAccount::Account.new(1, -1) }.to raise_error(ArgumentError)
+      end
+    end
+  end
```
Elffers, line 35: Excellent use of `raise_error` :+1: If you want to be extra fancy, you can also [test for the error message itself](https://www.relishapp.com/rspec/rspec-expectations/v/3-2/docs/built-in-matchers/raise-error-matcher#match-class-+-message-with-string):
`expect { BankAccount::Account.new(1, -1) }.to raise_error(ArgumentError, "insufficient opening balance!")`

lib/bankaccount/account.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      if amount <= @balance - min_balance
+        @balance -= amount
+      else
+        # may not overdraw; will output a warning and return un-modified balance [breakfast]
+        puts "Insufficient funds! You have $#{@balance} in your account."
+        @balance
+      end
+    end
+
```
Elffers, line 25: Your logic here is solid :+1:  You might want to consider extracting the check for sufficient funds into a separate method, so that the `withdraw` method itself only has a single responsibility (namely, deducting the funds).

lib/bankaccount/checkingaccount.rb
```diff
@@ -0,0 +1,35 @@
...
+    def withdraw(amount, transaction_fee = 1, min_balance = 0) # [lunch]
+      super
+    end
+
+    def withdraw_using_check(amount) # [lunch]
+      withdraw(amount, check_fee, -10) # (amount, transaction_fee, min_balance)
+
+      # increment @checks_used *after* check for sufficient funds
+      @checks_used += 1
+      return @balance
```
Elffers, line 19: Since Ruby returns the last line of a method by default, you don't need to say `return` explicitly--you can just say `@balance`.

Elffers:
Great work overall! You seem very comfortable with the concept of inheritance. I would definitely consider using constants rather than having default arguments, bc those can be accidentally mis-entered/overridden. 

One last thing to consider: for `CheckingAccount`, how might you handle checking if you've used the maximum number of free checks allotted per month? And can you find a gem that helps with testing time-sensitive method calls?

spec/lib/checkingaccount_spec.rb
```diff
@@ -0,0 +1,169 @@
+describe BankAccount::CheckingAccount do
```
Elffers, line 1: Note: If you wanted to avoid having to say `BankAccount::CheckingAccount` every time in your tests, you can wrap the entire file in `module BankAccount` (the same way you wrapped the class you're testing) and everything should be properly namespaced.

lib/bankaccount.rb
catchingash, line 9: Good to know. Is there a good time to actually use Pathname? (If it's something beyond my current level, I understand.)

lib/bankaccount/account.rb
catchingash, line 6: :+1:

lib/bankaccount/savingsaccount.rb
catchingash, line 19: Completely makes sense to me to not un-necessarily expose the variables via the params.

To an extent, if I wanted to take advantage of the inheritance, I still have to include "extra" things in the parent method though, right? Since Account doesn't actually have transaction fees or a minimum balance, it's just the children accounts that do?

lib/bankaccount/checkingaccount.rb
catchingash, line 33: The ternary operator is awesome! I keep forgetting it exists when I actually have a chance to use it, though, so it's not sticking in my head :(

spec/lib/account_spec.rb
catchingash, line 35: I wasn't sure what the best practice is here for that.... is it good to just have it pass on any error so that if someone decides to change the error message it doesn't break the test, or is it better to include the message so that we can check to make sure that they are getting the _correct_ error? Or should there be different tests for these?

lib/bankaccount/account.rb
catchingash, line 25: :+1:

lib/bankaccount/checkingaccount.rb
catchingash, line 19: :+1:

spec/lib/checkingaccount_spec.rb
catchingash, line 1: WOAH! Totally obvious, and yet totally mind-blowing all at once ;)

catchingash:
It would've made more sense to make a "free checks used?" type of method to extract the logic around the maximum number of free checks. If we had a database attached to this, I could look for check transactions within the current month (e.g. June) rather than incrementing an instance variable. The Time library and the Chronic gem both handle time (Chronic as a easier interface for non-standard inputs, letting me do a "this month" type of thing)... is that what you meant by time-sensitive method calls?

PR #20, opened 2015-06-01
completed Bank Accounts project

a-lmx:
I am submitting this for the dinner level.

I worked on making methods inheritable, particularly the #withdraw method. To do this, I moved some logic into a validate_withdrawal method, and changed that in inheriting classes when necessary. I feel like this could be worked out much better (clearer inheritance), and would appreciate comments towards that.

I also think I could have made more modular methods to pull out more functionality, but by the time I thought about that, I'd spent so much time working to get #withdraw inheritable that I ran out of time.


lib/bank_accounts/money_market_account.rb
```diff
@@ -0,0 +1,61 @@
+module BankAccounts
+
+	class MoneyMarketAccount < Account
```
GitOnUp, line 3: The readme.md is ambiguous about this (and this may be included in your comments of running out of time), but inheriting from SavingsAccount here helps you DRY and also gives you the functionality from the rest of Account.

lib/bank_accounts/checking_account.rb
```diff
@@ -0,0 +1,50 @@
...
+					valid = false
+					return @balance, valid
+				end
+			else
+				super
+			end
+		end
+
+		def withdraw_using_check(amount)
+			@using_check = true
```
GitOnUp: Since withdraws using checks are only done in checking accounts, the branch for `if @using_check` in validate_withdrawal could just live in this function and you could use the inherited validate_withdrawal.

This also helps avoid some pitfalls with `raise`: if validate_withdrawal or a deeper function raises an error, `@using_check` does not get reset to false on line 43.

GitOnUp:
Looks good to me overall.

> I moved some logic into a validate_withdrawal method, and changed that in inheriting classes when necessary. I feel like this could be worked out much better (clearer inheritance), and would appreciate comments towards that.

This is actually called the [template method pattern](http://en.wikipedia.org/wiki/Template_method_pattern) and is a totally reasonable way of doing this.

Good job!  Let me know if you have any other questions.

a-lmx:
Thanks for your comments, George. I like pulling the validation for check withdrawal out and not trying to make the single validate_withdrawal method do too much.

PR #21, opened 2015-06-01
Dinner (6/6)

amirahaile:
Whew! Sixty-four specs! I wanted to be really thorough with them this time around. I felt I could've easily missed opportunities for tests in Scrabble. Though I'm wondering if it's worth it to break up something like "CheckingAccount#deposit(amount) adds amount to the balance" and "CheckingAccount#deposit(amount) returns updated balance" since they both could be checked with `expect(account.deposit(5000)).to eq(15000)`…? I like that the separate tests would narrow down a potential problem, but I'm not sure if being that nitpicky is worth it.

PS – I know I sacrificed writing efficient code in order to finish Dinner. I know I didn't have to finish Dinner, but I really wanted to do it. Please point out any bloated code if you spot it. I'll work on balancing my priorities in future projects.


lib/bank/account.rb
```diff
@@ -0,0 +1,37 @@
+module Bank
+  class Account
+    # attr_accessor creates an implicit balance method
+    attr_accessor :id, :balance
```
GreyKn: As attr_accessors, id and balance are both read and writeable. In the context of a bank account, is that something you would want a user to be able to alter directly?

lib/bank/account.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def initialize(id, initial_balance)
+      raise ArgumentError, "Negative balance is invalid." if initial_balance.to_i < 0
+
+      @id = id
+      @balance = initial_balance.to_i
```
GreyKn, line 10: Something to be careful about is that money is allowed to be fractional, so using integers limits you to whole dollars, so  you may get some funky behavior later.

lib/bank/account.rb
```diff
@@ -0,0 +1,37 @@
...
+      @minimum_balance = 0
+    end
+
+    def withdraw(amount)
+      # Integer() will turn strings of numbers into valid integers.
+      # If it's not a whole number or includes invalid characters
+      # it'll return 0 instead
+      amount = Integer(amount) rescue 0
```
GreyKn: Using `Integer()` is a good way of dealing with parsing numbers into strings. I'm a little concerned about the rescue, because you're triggering something known as a silent failure.

So if someone tries to `withdraw('eleventy billion dollars')`, this will fail, default the withdraw amount to 0, and return the original balance. But there is no indication on why the withdrawal failed. So in this case, an `ArgumentError` would be appropriate:

```
begin
  amount = Integer(amount)
rescue
  raise ArgumentError, 'Withdraw amount not a number'
end
```

lib/bank/account.rb
```diff
@@ -0,0 +1,37 @@
...
+        @balance -= amount
+      else
+        @is_withdrawal_error = true
+        puts error_message
+      end
+
+      return @balance
+    end
+
+    def deposit(amount)
```
GreyKn, line 37: Same issue as with `#withdraw` with silent errors

lib/bank/checking_account.rb
```diff
@@ -0,0 +1,44 @@
...
+        # subtract the $1 fee from balance
+        return @balance - 1
+      else
+        # return the unmodified balance
+        return @balance
+      end
+    end
+
+    def withdraw_using_check(amount)
+      if @month != Time.now.mon
```
GreyKn, line 24: Great bonus feature of resetting your check count :100:

lib/bank/checking_account.rb
```diff
@@ -0,0 +1,44 @@
...
+    attr_accessor :check_transactions, :month
+
+    def initialize(id, initial_balance)
+      super
+      @check_transactions = 0
+      @month = Time.now.mon
+    end
+
+    def withdraw(amount)
+      super
```
GreyKn, line 12: Wow, this is a really great way to use `#super` to be DRY

lib/bank/checking_account.rb
```diff
@@ -0,0 +1,44 @@
...
+      # there was not an error with the withdrawal amount
+      if !@is_withdrawal_error
+        # subtract the $1 fee from balance
+        return @balance - 1
```
GreyKn: You probably want this to be 

```
return @balance -= 1
```

Otherwise, while the method seems to return the right balance, the actual balance will be unchanged (fee not assessed).

lib/bank/checking_account.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      # adds $2 fee if over 3 free check limit
+      amount += 2 if @check_transactions >= 3
+
+      unless @balance - amount < -10
+        @check_transactions += 1
+        return @balance -= amount
+      else
+        puts "Account cannot overdraft over $10."
```
GreyKn, line 36: Should you return the balance amount even if `#withdraw_using_check` fails? That would be consistent with the existing `#withdraw` method

lib/bank/money_market_account.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def initialize(id, initial_balance)
+      raise ArgumentError, "Account requires minimum balance of $10,000." if initial_balance.to_i < 10_000
+      super
+      @num_of_transactions = 0
+      @balance = initial_balance.to_i
+      @month = Time.now.mon
+    end
+
+    def is_at_transaction_limit
```
GreyKn: The name of this method sets the expectation that it should return a boolean. So think about naming it something like `#check_transaction_limit`

lib/bank/money_market_account.rb
```diff
@@ -0,0 +1,69 @@
...
+      if !@is_deposit_error && balance_before_deposit > 10_000
+        @num_of_transactions += 1
+      end
+
+      return @balance
+    end
+
+    def add_interest(rate)
+      @interest = @balance * rate / 100
+      @balance += @interest
```
GreyKn, line 65: According to the assignment, while you're correct to update balance with added interest, the actual return should be the interest amount, not the updated balance

lib/bank/savings_account.rb
```diff
@@ -0,0 +1,30 @@
+module Bank
+  class SavingsAccount < Account
+
+    def initialize(id, initial_balance)
+      raise ArgumentError, "Balance must be a minimum of $10." if initial_balance < 10
+
+      @id = id
+      @balance = initial_balance.to_i
+      @minimum_balance = 12
```
GreyKn: The minimum balance should be $10, I'm wondering why you have it set to $12?

lib/bank/savings_account.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def withdraw(amount)
+      super
+      # there was not an error with the withdrawal amount
+      if !@is_withdrawal_error
+        # subtract the $2 fee from balance
+        return @balance - 2
```
GreyKn: Same as CheckingAccount, this returns what looks like an updated balance, but the fee is not actually accessed on the transaction.

lib/bank/savings_account.rb
```diff
@@ -0,0 +1,30 @@
...
+      else
+        # return the unmodified balance
+        return @balance
+      end
+    end
+
+    def add_interest(rate)
```
GreyKn: I see that you're repeating `#add_interest` in both Money Market and Savings Accounts. Can you think of a way you could only write it once?

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+describe Bank::Account do
+
+  before :each do
+    @bank = Bank::Account.new("Checking", 50)
+  end
+
+  describe "#new" do
+    it "creates a new instance" do
+      expect(Bank::Account).to respond_to(:new)
```
GreyKn: All Ruby classes inherit from Object, which defines `.new`, so you don't need to really test for it.

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
+require 'bank'
+
+describe Bank::Account do
+
+  before :each do
+    @bank = Bank::Account.new("Checking", 50)
+  end
+
+  describe "#new" do
```
GreyKn, line 7: Good consistent use of the describe keyword :+1:

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
+require 'bank'
+
+describe Bank::Account do
+
+  before :each do
+    @bank = Bank::Account.new("Checking", 50)
```
GreyKn: You can simplify this by using `let` instead of a `before` block

```
   let(:bank) { Bank::Account.new("Checking", 50) }
```

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+    describe "initialize" do
+      it "assigns the id instance variable" do
+        expect(@bank.id).to eq("Checking")
+      end
+
+      it "assigns the balance instance variable" do
+        expect(@bank.balance).to eq(50)
+      end
+
+      it "sets balance to an integer" do
```
GreyKn, line 17: In this case, you should split these into 2 tests, once for accepting a number, and once for accepting a nonnumber but still having balance 0.

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+          expect{Bank::Account.new("Checking", -10)}.to raise_error(ArgumentError, "Negative balance is invalid.")
+        end
+      end
+    end
+  end
+
+  describe "#withdraw(amount)" do
+
+    it "removes amount from the account" do
+      [45, 25, 10, 0.10].each do |money|
```
GreyKn, line 38: :+1: for combining arrays and a loop to test a range of withdraw values

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+        bank.withdraw(money)
+
+        expect(bank.balance).to eq(1000 - money.to_i)
+      end
+    end
+
+    # it "is a result of an ATM transaction" do
+    # end
+
+    it "subtracts an integer from @balance" do
```
GreyKn, line 46: Same as in your initializer tests, you should split this into 2 tests, one for a number and another for a nonnumber

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+        @bank.withdraw(100)
+        expect(@bank.balance).to eq(50)
+      end
+
+      it "outputs a warning message" do
+        expect{@bank.withdraw(100)}.to output("You cannot withdraw more than the balance minimum of $0.\n").to_stdout
+      end
+
+      it "returns unmodified balance" do
+        expect(@bank.withdraw(100)).to eq(50)
```
GreyKn: This assumes some magic number, so what you could do to make this test less brittle is:

```
unmodified_balance = @bank.balance
expect(@bank.withdraw(100)).to eq(unmodified_balance)
```

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      end
+    end
+  end
+
+  describe "#deposit(amount)" do
+    money = ["0.30," "50", "60"].shuffle.sample
```
GreyKn: `#sample` already returns a random element, so there's no need to `#shuffle` beforehand

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    it "adds amount to @balance" do
+      @bank.deposit(money)
+      expect(@bank.balance).to eq(50 + money.to_i)
```
GreyKn: I think this test. This is similar to a concept known as **fuzzing**, where you use varying, potentially unexpected, values to check for a break

spec/lib/bank/account_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+    # as a result of an ATM transaction
+
+    it "returns updated balance" do
+      expect(@bank.deposit(money)).to eq(50 + money.to_i)
+    end
+
+    context "adding invalid amounts to balance" do
+      it "only adds integers" do
+        ["af", "$$", "5&3"].each do |fake_money|
+          expect(@bank.deposit(fake_money)).to eq(50 + 0)
```
GreyKn: Why have the `+ 0` in your expectation?

spec/lib/bank/checking-account_spec.rb
```diff
@@ -0,0 +1,138 @@
+describe Bank::CheckingAccount do
+  it "creates a new instance" do
+    expect(Bank::CheckingAccount).to respond_to(:new)
+  end
+
+  it "inherits from Bank::Account" do
+    expect(Bank::CheckingAccount.superclass).to eq(Bank::Account)
```
GreyKn, line 7: Interesting sanity check on your class hierarchy :+1:

spec/lib/bank/checking-account_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+    it "withdraws amount from balance" do
+      account.withdraw(100)
+      expect(account.balance).to eq(900)
+    end
+
+    # from an ATM transaction
+
+    # how can I check for the $1 independently?
+    it "also withdraws $1 fee and returns balance" do
+      expect(account.withdraw(100)).to eq(899)
```
GreyKn: So one way you can write this test is like this:

```
old_balance = account.balance
fee = 1 #If you had a reader for fee, it would help here
withdraw_amount = 100
new_balance = old_balance - withdraw_amount - fee

expect(account.withdraw(withdraw_amount)).to eq(new_balance)
expect(account.balance).to eq(new_balance)
```

lib/bank/account.rb
```diff
@@ -0,0 +1,37 @@
...
+      # it'll return 0 instead
+      amount = Integer(amount) rescue 0
+      error_message = "You cannot withdraw more than the balance minimum of $#{@minimum_balance}."
+      @is_withdrawal_error = false
+
+      if @balance - amount >= @minimum_balance
+        @balance -= amount
+      else
+        @is_withdrawal_error = true
+        puts error_message
```
GreyKn: Consider using `warn` instead of `puts`, it's basically the same thing here, but contextually it's clearer the the user what's going on.

spec/lib/bank/checking-account_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+
+        3.times do
+          @account.withdraw_using_check(20)
+          # @account.balance = 940
+        end
+      end
+
+      it "charges a $2 fee per use after the third check transaction" do
+        @account.withdraw_using_check(20)
+        # 940 - (20 + 2) = 918
```
GreyKn: Good comment to give context to the reader about what the test is doing. Even better would be to actually use variables in your test (check above example)

spec/lib/bank/checking-account_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      expect(account.check_transactions).to eq(5)
+    end
+
+    context "when three checks have been used this month" do
+      # used before block to solve scope issue
+      # apparently :account can't be used within a context block
+      # outside of the individual examples (e.g. it blocks)…?
+      before(:each) do
+        @account = Bank::CheckingAccount.new("Checking", 1000)
```
GreyKn: A variable defined by `let` should be usable in a `before` block, you just need to define the `let` before the `before`.

GreyKn:
Hi Amira,

Great job on completing the dinner level of this assignment. I was really happy with the way you were using `super` in places to take advantage of inheritance (thus writing less code), and I feel your specs were pretty spot on.

There were a couple of misunderstandings with the project specifications on return values, and I had a couple suggestions for your tests, but overall, I think you successfully completed Dinner. :tada: 

Let me know if you have any questions.

-Chef

lib/bank/account.rb
amirahaile, line 10: I thought about that… since most of these methods are responding to ATM transactions, I'm assuming you would only deposit/withdraw whole dollars. I'll double check that my check transaction methods allow for this though.

lib/bank/savings_account.rb
amirahaile: I did it as a quick way to make sure the withdrawal could be made (including the $2 fee) but I rewrote how this was handle and like the new solution much better. Thanks!

lib/bank/savings_account.rb
amirahaile: I tried to use modules and mixins but to no avail. How can this be done?

amirahaile:
Thanks for the helpful tips! I learned a lot fixing this up. In one of your comments you suggested I use `warn` instead of `puts` and I liked that a lot. However, I don't know how to expect it in my spec tests. I found the deprecated way of using `should` but couldn't find its `expect` equivalent.

PR #23, opened 2015-06-01
Completed BankAccounts project

blip-lorist:
Finished Lunch - turns out the "month" thing isn't tied to Time. (Lines 34 - 44 in checking_account.rb are just extra untested code, then!). Overall, I really enjoyed this project, got some very encouraging feedback from a TA who told me my breakfast level was "solid code." :D 

I also got a little stuck on Rspec for the lunch level trying to run a method 3 times within rspec - will probably email questions if I don't figure it out by Monday.


kariabancroft:
I would recommend removing the RSpec coverage folder from your git checkin via the .gitignore file

lib/bank_account/account.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    # Puts monies into account, updates balance
+    def deposit(amount)
+      @current_balance += amount
+    end
+
+    # Returns balance
+    def balance
+      @current_balance
```
kariabancroft: If you used an instance variable of `@balance` instead of `@current_balance` then you wouldn't need a separate method for this you could've just used the `attr_reader`.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # Get monies
+    def withdraw(amount)
+
+      # Proceeds if you have enough monies
+      if amount <= @current_balance
+        @current_balance -= amount
+      else
+        # Yells at you (nicely) when you don't have enough monies
+        return "Sorry, you don't have that much! Withdraw canceled. Let's keep your current balance at #{@current_balance}."
```
kariabancroft: I think the requirements want you to return the original balance rather than returning the error message.

lib/bank_account/account.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  class Account
+    # Makes id and initial balance viewable
+    attr_reader :id, :initial_balance
+    # Makes current balance changable
+    attr_accessor :current_balance
+
+    def initialize(id, initial_balance)
+      @id = id
+      @initial_balance = initial_balance
```
kariabancroft: Do you need the `@initial_balance` isntance variable for use later or do you only need one instance variable to keep track of the balance?

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,64 @@
...
+    def initialize(id, initial_balance)
+      super
+      @free_checks = 3 #Per month
+    end
+
+    # Get monies
+    def withdraw(amount)
+
+      slight_burglary = 1 #That yucky fee
+      total_withdraw = amount + slight_burglary
```
kariabancroft, line 19: This is a nice use of a local variable since this calculation is needed more than once

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,64 @@
...
+      # Resetting @free_checks to 3 each subsequent month
+      # Grabs the calendar month when a check is first used for withdraw
+      first_time_use = Time.new.month
+      # Sets an "alarm" for next month
+      next_month = first_time_use + 1
+      # Runs a check
+      if Time.now.month == next_month
+        @free_checks = 3
+      end
+      # I don't think this works the way I want it to.
```
kariabancroft, line 43: I like that you attempted to use the time functionality to reset the check amounts! I think you would actually need some sort of ongoing "job" that would reset the checks every month. Another option would be keeping track of the # of days that have elapsed between the last time it was reset.

lib/bank_account/checking_account.rb
```diff
@@ -0,0 +1,64 @@
...
+      # Reduces free checks by one, complete the transaction
+      if @current_balance - amount >= -10 && free_checks > 0
+        @free_checks -= 1
+        @current_balance -= amount
+      # Adds that yucky fee when there are no free checks
+      elsif @current_balance - amount >= -10 && free_checks == 0
+        @current_balance -= total_withdraw
+      # Yells at you (nicely) when your balance goes under -$10
+      else
+        return "Sorry, you can't overdraft more than $10! Withdraw canceled. Let's keep your current balance at #{@current_balance}."
```
kariabancroft: Same case here - I think you want to return the `@current_balance` rather than the error message

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,41 @@
...
+
+
+  class SavingsAccount < Account
+
+    def initialize(id, initial_balance)
+
+      #Inherit the same variables as Account
+      # I realize that I'll also inherit the same negative balance
+      # conditional. However, that's intentional.
+      super
```
kariabancroft, line 11: Nice use of the base classes initialize functionality here

lib/bank_account/savings_account.rb
```diff
@@ -0,0 +1,41 @@
...
+        @current_balance -= total_withdraw
+      else
+        # Yells at you (nicely) if you don't have enough monies
+        return "Sorry, you need a minimum of ten whole bucks in your account. Withdraw canceled. Let's keep your current balance at #{@current_balance}."
+      end
+    end
+
+    # Calculates and displays the interest
+    def add_interest(rate)
+      interest = @current_balance * (rate/100)
```
kariabancroft: Is this actually adding the interest to the balance or just returning the interest?

spec/lib/bank_account/savings_account_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it "won't let you dip under $10" do
+      expect(new_savings_account.withdraw(195)).to eq("Sorry, you need a minimum of ten whole bucks in your account. Withdraw canceled. Let's keep your current balance at 200.")
+    end
+
+    it "lets you deposit moneh" do
+      expect(new_savings_account.deposit(5)).to eq(205)
+    end
+
+    it "lets you check interest" do
+      expect(new_savings_account.add_interest(0.25)).to eq("$0.5")
```
kariabancroft, line 38: If you check the balance after adding the interest in your test - you might realize that the interest wasn't added. I'd recommend updating this test when/if you update the add interest method.

kariabancroft:
Nice job on this! I'm glad that you ran the RSpec coverage report. There are only a few small items to clean up, but overall your code is nice and clear and concise. I like that you tried to use the time to reset the checks, there will be good opportunities to do things like this on later assignments.

lib/bank_account/account.rb
lorainekv: Oh ya. For some reason, I got mixed up thinking that I needed it in attr_accessor to update it. Created a setter method with @balance instead.

lorainekv:
Yay, I was able to update everything. Consolidated everything in the balance instance variable, instead of having current_balance and initial_balance throughout the project. Updated all specs.

PR #24, opened 2015-06-01
Bank Account (Breakfast and part of Lunch) - Corinne Pingul

corinnepingul:
I finished all the breakfast requirements for this project and started the lunch requirements, but only finished the first method #withdraw(amount). One thing that I was a bit frustrated with was trying to figure out how to inherit the #withdraw(amount) from the Account class so I didn't have to re-write the code over again in each class. My solution was to create constant variables for each class that they could use to account for the transaction fees and minimum balance requirements for each account.


lib/banks/account.rb
```diff
@@ -0,0 +1,33 @@
+module BankAccount
+  class Account
+
+    def initialize(id, initial_balance)
+      @id = id
+      @current_balance = initial_balance
+
+      if initial_balance <  0
```
jnf, line 8: Generally, we'd want to see these kinds of validations _before_ we did any 'work' in the method. Assigning `@current_balance` is harmless enough, probably, but you can't guarantee that there aren't any side effects. In the future, we'll do things like build separate objects/processes for validation and prevent even object instantiation until we know we have good input.

lib/banks/account.rb
```diff
@@ -0,0 +1,33 @@
...
+        # accounts cannot be created with a negative balance
+        raise ArgumentError.new("Bank accounts cannot be opened with a negative balance.")
+      end
+    end
+
+    def balance
+      return @current_balance
```
jnf, line 15: Hey hey, this could be an `attr_reader`, yeah?

lib/banks/checking.rb
```diff
@@ -0,0 +1,17 @@
...
+module BankAccount
+  class CheckingAccount < Account
+      MINIMUM_BALANCE = 0
+      TRANSACTION_FEE = 1
+
+    def withdraw(amount)
+      # This method inherits from the Account class #withdraw(amount) method
+      # amount will be checked against current_balance to make sure you have
+      # enough money in your account to make this withdrawl (leaving at least
+      # $10 left in the account)
```
jnf, line 11: Minor gripe: updating comments after copypasta. `MINIMUM_BALANCE` is zero in this class.

lib/banks/checking.rb
```diff
@@ -0,0 +1,17 @@
...
+      amount = amount + MINIMUM_BALANCE + TRANSACTION_FEE
+      # need to add MINIMUM_BALANCE back on the current_balance
+      super + MINIMUM_BALANCE
```
jnf, line 14: I was pretty confused when I saw this method the first time, but I dig what you're doing. You're manipulating the amount of the withdrawal before invoking the inherited behavior. It's kinda confusing to me because when a method invokes `super` without parens/parameters, `super` gets the same parameters as was given the inheriting method.

I read line 14 as `super` would receive the `amount` parameter _as it was passed into the `#withdraw` method_, but that's not exactly right. You mutate `amount` on line 12, and that mutated value is passed into super. Just for clarity, I think I'd write this like...

``` ruby
adjusted_amount = amount + MINIMUM_BALANCE + TRANSACTION_FEE
super(adjusted_amount) + MINIMUM_BALANCE
```

That'd be more clear as to what you're giving `super` and why. All that aside, your commenting is great here and helped me understand your process.

lib/banks/savings.rb
```diff
@@ -0,0 +1,39 @@
...
+      # This method inherits from the Account class #withdraw(amount) method
+      # amount will be checked against current_balance to make sure you have
+      # enough money in your account to make this withdrawl (leaving at least
+      # $10 left in the account)
+      amount = amount + MINIMUM_BALANCE + TRANSACTION_FEE
+      # need to add MINIMUM_BALANCE back on the current_balance
+      super + MINIMUM_BALANCE
+    end
+
+    def rate_as_a_percent(rate)
```
jnf, line 23: zomg this method makes me so happy. Really well done with this. You've identified an edge case; isolated its handling in a reusable method, and provided comments to clarify your intent. :100: A+++++, would review again.

jnf:
Great job on this project. Your specs are very clear and well organized. I left a couple of comments; have a look and let me know if you've any questions or concerns. :)

spec/savings_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    it "does not allow the account to go below $10 - outputs warning and returns previous balance" do
+      expect{savings_account.withdraw(87)}.to output("WARNING: You cannot remove that much money from your account.\n").to_stdout
+      expect(savings_account.withdraw(87)).to eq(98) # the previous balance is 98 because of the last test (we withdrew 100)
+    end
+  end
+
+  context "Adding interest" do
+    savings_account = BankAccount::SavingsAccount.new(45, 10000)
```
jnf, line 37: It'd prolly be best to wrap this kind of logic in a `let`, like:

``` ruby
let(:savings_account) { BankAccount::SavingsAccount.new(45, 10000) }
```

`let` is rad, and [here's why](http://stackoverflow.com/questions/5359558/when-to-use-rspec-let/5359979#5359979).

lib/banks/account.rb
corinnepingul, line 15: Oh, yeah 👍

lib/banks/checking.rb
corinnepingul, line 11: Whoops!

lib/banks/checking.rb
corinnepingul, line 14: Oh, okay yeah I can see how that is super unclear as it is written.

spec/savings_spec.rb
corinnepingul, line 37: Yes! I remember reading about `let` during this project, but I'm not sure why I didn't use it here...whoops!

corinnepingul:
Thanks Jeremy for the comments!

